PR #22, opened 2024-03-24
feat: implement authentication

loggerz:
Todo:

- [x] Sign in using Supabase API with Google as Identity Provider
- [x] Logout flow to enable users to switch accounts
- [x] Google authentication bypass for other platforms with a fake account
- [x] State-based routing ? Or at least : UI should handle logout
- [x] Refactor state management, need simple access to basic user metadata
  
  Related documentation:
- [Official tutorial for "Simple" state management](https://docs.flutter.dev/data-and-backend/state-mgmt/simple#providerof)
- [Navigation & routing](https://docs.flutter.dev/cookbook/navigation/navigation-basics) - maybe use this instead of conditional rendering with global `ChangeNotifiy`


loggerz:
Finally wrapped my head around this little maze, recap of this part :
- Google Sign In package is wrapped with Supabase authentication api
- Alternative login method is available for easier testing on non-android platforms (web/linux)
- The authentication state is reactive and immediately updated with supabase events
- App state is implemented with the Provider API, recommended by flutter as a sane default
- Data from state is easily obtainable with the Provider API, there might be too much widget re-draws but let's not pre-optimize before noticing a performance hit

ClementJosso:
There appears to be an issue when launching on a web device, specifically in the login view, which gives the error "Unsupported operation: Platform._operatingSystem".

pmdlt:
Same here 
image

loggerz:
Ok, this bug is explained by this upstream issue : https://github.com/flutter/flutter/issues/36126#issuecomment-511145268

Fixed according to their recommendation (using another API that's compatible with web platform)

pmdlt (dismissed):
All good on web (Google not tested)

Pacsort17:
My merge request was an error.

Test on an android -> Fake login works, but Real login don't update the view (it stays in the first screen), however we can select our google account

Pacsort17 (changes requested):
My merge request was an error.

Test on an android -> Fake login works, but Real login don't update the view (it stays in the first screen), however we can select our google account

Logging with a real google account should lead to the Logined Welcome Page

ClementJosso (changes requested):
On web run : Fake account works well with logged in view correctly displayed.
On Android emulator run : Can login with fake account as well, but signing in with google always goes back to login view even after selecting an account. Never access the logged in view.

loggerz:
The relevant part of the logs you provided is `PlatformException(sign_in_failed, com.google.android.gms.common.api.ApiException: 10: , null, null)"` . This indicates a bad configuration of google sign in. 

I suspect the debug signing certificate is different for each one of us thus preventing the use of the current google oauth consent screen configured with my hash, looking into a better way of handling this...

loggerz:
I commited the debug keystore I used to configure the oauth consent screen and updated gradle to use it. This last issue should now be fixed

ClementJosso (dismissed):
Test on Android Emulator -> Works well, can log in and log out of multiple Google accounts, as well as the fake account.

pmdlt (approved):

Pacsort17 (approved):
Runs on edge.
Runs on Android and works for my private and EPFL account.
Well done :)

PR #33, opened 2024-03-28
feat: add a dummy news page with correct UI

ClementJosso:
This should be a working news page, it just needs the viewmodel to get the actual data of the user.

To test it, you can edit line 40 of main.dart to have : "body = const NewsView();" (and import news_view.dart)


Thechi2000 (commented):
Just a nit pick, otherwise good work !

Pacsort17 (changes requested):
Should be changed :
Tested on edge and android : the overlay when passing the mouse / finger over the news is contained in the text instead of surrounding it (the text should be in the pink rectangle, not the other way around).

Works :
The view is scrollable and detects when it's clicked, outputting a source.

Pacsort17 (approved):
Overlay for sources is not there anymore (except pour AppBar). Clicking on the text still gives the source (only text 4 outputs "source2"), and clicking somewhere else (like titles) doesn't output anything.

Thechi2000 (approved):
Only tested on Linux, not Android. Looks great !

PR #35, opened 2024-04-06
28 dev profile page data supabase integration

pmdlt:
As discussed at our last meeting, we add this model and viewmodel to the main branch, so that it can be added to the view.
Thanks for your review (which I'm sure will be very useful) !


Thechi2000 (approved):
Looks good to me !

PR #41, opened 2024-04-10
40 test coverage

Thechi2000:


Pacsort17 (commented):
How can we have the coverage locally ? Maybe on windows it doesn't work :')

Thechi2000:
> How can we have the coverage locally ? Maybe on windows it doesn't work :')

You need the lcov toolchain, I think this [should](https://github.com/valbok/lcov) work. Then, the `test.sh` script shows the commands to run to generate the coverage report.

LeanderCretegny (approved):
works nicely

PR #49, opened 2024-04-11
38 feat authentication UI

quartztz:
Clean up and get UI closer to Figma. Behavior for guest sign in can be tested in chrome, behavior for google sign in works best on devices. 

Please review!!


quartztz:
As always when introducing a new platform, sonarcloud complains... Best practice would have me create a new PR to fix its config, but having discussed irl with the team, it's going to be faster and more relevant to just change the config on this branch and approve both changes as well. So that's that!

Thechi2000 (approved):
Great job ! Tested on Linux and Android.

Just a small question: why did you need to change the macos specific files ?

Thechi2000:
@quartztz It's perfectly fine, I was just wondering if there was a particular reason (e.g. the auth did not work on mac).

quartztz:
Oh no, I think it's just flutter behavior. Still untested on `macos`, but works on android!

loggerz (changes requested):
We should not add another platform as some features need specific implementation. Main platform is android and web + linux are used for testing purposes

quartztz:
Oh! I see. I'll remove them from the repo.

Thechi2000 (approved):

loggerz (approved):

PR #55, opened 2024-04-12
30 feat view profile page UI

Pacsort17:
Profile page UI done.

The page should look like the [figma](https://www.figma.com/file/8GfgXznK0gmJuPWhfBwGBs/ActualIA%3A%3Amockup?type=design&node-id=0%3A1&mode=design&t=iUov9lFi8X6xcy4T-1).

The username should be the email of the connected user.
The view should be scrollable.
Clicking on buttons should print their name in the console, except for the logout that redirect to login page.

I tested it on Android and Edge.


Thechi2000 (approved):

ClementJosso (approved):
Tested on chrome, works as intended

PR #56, opened 2024-04-12
54 feat combine transcript cloud function and news view model

ClementJosso:
Needs the actual cloud function to be implemented but otherwise should work well.
To test it : Change line 55 in main to replace the WizardView by NewsView.
Expected output: If tested with guest account on 12/04, should display a news
If other account or other day, should load 5 seconds then tell you something went wrong while generating news.


LeanderCretegny (approved):

quartztz (approved):
looks great! thanks for the work :D

PR #77, opened 2024-04-16
76 feat add audio generation

pmdlt:
This PR introduces functionality to generate audio files from transcripts stored in the database.

**Changes:**

Added a Deno server to handle HTTP requests.
Integrated Supabase for database operations.
Integrated the OpenAI API for text-to-speech functionality.
Implemented error handling for various scenarios.
Added support for specifying desired voice options.
Uploaded generated audio files to Supabase storage.
Updated database records with audio file paths.

**Purpose:**

The purpose of this PR is to automate the process of generating audio files from text transcripts.


ClementJosso (dismissed):
Great work!

Thechi2000 (changes requested):
Works great, thanks !

I think we should have some tests, at least to cover errors for bad requests (missing parameters and so on). Since it is usable with the anon key, it is also possible to have tests for working case, but this may slow the CI (to discuss).

Thechi2000 (approved):

ClementJosso (approved):

PR #83, opened 2024-04-18
82 update news view model and model to store source and audio info

ClementJosso:
This PR updates the model and view model of the news to allow storage of the audio file info as well as the source info.
When launching the app, it should work exactly as it did before, except that clicking on a paragraph now prints the name of the newspaper/webiste that made the article and not a URL.

### DIFF
--- a/lib/models/news.dart
+++ b/lib/models/news.dart
@@ -1,18 +1,31 @@
 class News {
   final String title;
   final String date;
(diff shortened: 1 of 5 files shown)


Pacsort17 (approved):
Tested on android : Nothing changed from last time, except the excpected prints.

PR #86, opened 2024-04-18
Offline Recorder

Pacsort17:
API to store news locally in ./offlineNews/ 
Constructor creates the folder, and some functions allows to store files in it.

TODO : How do we implement the offline ? Every news is stored, if yes how do we manage space ?

TODO : Create tests


Thechi2000:
For the space management, the user should be able to at least view the used space, and maybe change the maximum size the app is allowed to use. This would happen in the settings view. Then, the app would drop the least recent news to match this limit.

We may want to split this into two features, to discuss.

pmdlt:
(to update changes in files)

pmdlt:
There are several tests that do not work yet, I take the liberty of putting this PR in draft so as not to be disturbed by review notifications on the Telegram group :)

loggerz:
Can we have instructions for testing the feature ? I'm unable to load news in airplane mode (it's loading indefinitely without error)

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+    final file = File(filePath);
+
+    // Cannot have two transcripts for the same day
+    if (await file.exists()) {
+      await file.delete();
+    }
+
+    await file.writeAsString(json);
+
+    // TODO : Only delete oldest files when overflowing
```
Thechi2000, line 125: Does it need to be done in this PR, or is it a note for later ? If yes, it would be a good idea to open an issue for it (even before the PR is merged).

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+    if (await _dirSize(Directory(_appOfflineNewsPath)) >= _maxStorageSize) {
+      _cleanStorage();
+    }
+  }
+
+  /// Load data from local storage. Date must be formated in the following way : <year>-<month>-<day>
+  Future<News> loadNews(String date) async {
+    RegExp regex = RegExp(r"^[0-9]{4}-[0-9]{2}-[0-9]{2}$");
+    if (regex.allMatches(date).isEmpty) {
+      throw UnimplementedError("Date String not well formated");
```
Thechi2000, line 135:
```suggestion
      throw FormatException("Date String not well formated");
```

`UnimplementedError` is not adapted to this case, I think `FormatException` suits more.

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+          content: p["content"]);
+    }).toList();
+
+    return News(
+        title: decodedJson["title"],
+        date: decodedJson["date"],
+        transcriptId: decodedJson["transcriptID"],
+        audio: decodedJson["audio"],
+        paragraphs: paragraphs);
+  }
```
Thechi2000, line 163: It would be safer and more concise to use the functions generated in `lib/models/news.g.dart`.

test/unit/news_view_model.dart
```diff
@@ -105,10 +110,41 @@ class AlreadyExistingNewsVM extends NewsViewModel {
...
+    // TODO: implement loadNews
+    throw UnimplementedError();
+  }
+
+  @override
+  Future<int> setMaxStorageSize(int newStorageSize) async {
+    // TODO: implement setMaxStorageSize
+    throw UnimplementedError();
+  }
+}
```
Thechi2000, line 136:
```suggestion
class MockOfflineRecorder extends Fake implements OfflineRecorder {
  @override
  Future<void> downloadNews(News news) async {
    //does not need to do anything
  }
}
```

Using `extends Fake` will automatically replace all unimplemented methods with a stub throwing `UnimplementedError`.

test/unit/offline_recorder_mocks/mock_directory.dart
```diff
@@ -0,0 +1,133 @@
...
+  @override
+  // TODO: implement uri
+  Uri get uri => throw UnimplementedError();
+
+  @override
+  Stream<FileSystemEvent> watch(
+      {int events = FileSystemEvent.all, bool recursive = false}) {
+    // TODO: implement watch
+    throw UnimplementedError();
+  }
```
Thechi2000, line 132: As said above, using `extends Fake` will automatically replace all unimplemented methods with a stub throwing UnimplementedError.

test/unit/offline_recorder_mocks/mock_file.dart
```diff
@@ -0,0 +1,306 @@
...
+    // TODO: implement watch
+    throw UnimplementedError();
+  }
+
+  @override
+  Future<File> writeAsBytes(List<int> bytes,
+      {FileMode mode = FileMode.write, bool flush = false}) {
+    // TODO: implement writeAsBytes
+    throw UnimplementedError();
+  }
```
Thechi2000, line 245: As said above, using `extends Fake` will automatically replace all unimplemented methods with a stub throwing UnimplementedError.

test/unit/offline_recorder_mocks/mock_file.dart
```diff
@@ -0,0 +1,306 @@
...
+  @override
+  // TODO: implement modified
+  DateTime get modified => throw UnimplementedError();
+
+  @override
+  int get size => _size;
+
+  @override
+  // TODO: implement type
+  FileSystemEntityType get type => throw UnimplementedError();
```
Thechi2000, line 305: As said above, using `extends Fake` will automatically replace all unimplemented methods with a stub throwing UnimplementedError.

lib/models/news.dart
```diff
@@ -12,8 +17,35 @@ class News {
...
+  factory News.fromJson(dynamic json) {
+    return News(
+        title: json['title'] as String,
+        date: json['date'] as String,
+        transcriptId: json['transcriptID'] as int,
+        audio: json['audio'] as String?,
+        paragraphs: (json['paragraphs'] != null)
+            ? (json["paragraphs"]) as List<Paragraph>
+            : List.empty());
+  }
```
Thechi2000, line 30:
```suggestion
  factory News.fromJson(dynamic json) => _$NewsFromJson(json);
```

Can be done using the generated function.

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+import 'package:actualia/viewmodels/news.dart';
+
+/// Records news locally to retrieve them offline. Must be created through OfflineRecorder.create(newsProvider)
+class OfflineRecorder {
+  static const ROOT_OFFLINE_FOLDER = "storage";
+
+  late final _appOfflineNewsPath;
+  var _maxStorageSize = 17e+6.toInt(); // One minute mp3 file ~ 1MB
+
+  OfflineRecorder._create(String appOfflineNewsPath) {
```
Thechi2000, line 16: Why does it need a private constructor ?

test/unit/offline_recorder_mocks/mock_filesystem.dart
```diff
@@ -0,0 +1,160 @@
...
+    while (exist && depth < path.length) {
+      if (subFiles.containsKey(path[depth])) {
+        subFiles = subFiles[path[depth]];
+        depth++;
+      } else {
+        exist = false;
+        subFiles[path[depth]] = createSubDir(path.sublist(depth), depth++);
+      }
+    }
+    // debugPrint("[ADDDIR] files: $FILES");
```
Thechi2000, line 68:
```suggestion
```

Commented print should be removed imo.

test/unit/offline_recorder_mocks/mock_filesystem.dart
```diff
@@ -0,0 +1,160 @@
...
+        } else {
+          subFiles = subFiles[path[depth]];
+        }
+        depth++;
+      } else {
+        exist = false;
+        subFiles[path[depth]] = createSubDir(path.sublist(depth), depth++);
+      }
+    }
+    // debugPrint("[ADDFILE] files: $FILES");
```
Thechi2000, line 102:
```suggestion
```

Commented print should be removed imo.

test/unit/offline_recorder_mocks/mock_filesystem.dart
```diff
@@ -0,0 +1,160 @@
...
+        i++;
+      } else {
+        exist = false;
+      }
+    }
+    return exist;
+  }
+
+  bool dirExist(MockDir dir) {
+    // debugPrint("[DIREXIST] dir: $dir");
```
Thechi2000, line 129:
```suggestion
```

Commented print should be removed imo.

test/unit/offline_recorder_mocks/mock_filesystem.dart
```diff
@@ -0,0 +1,160 @@
...
+    List<String> path = dir.path.split("/");
+    return pathExist(path);
+  }
+
+  bool fileExist(MockFile file) {
+    // debugPrint("[FILEEXIST] file: $file");
```
Thechi2000, line 135:
```suggestion
```

Commented print should be removed imo.

lib/viewmodels/news.dart
```diff
@@ -26,7 +34,15 @@ class NewsViewModel extends ChangeNotifier {
     _newsList = newsList;
   }
 
-  NewsViewModel(this.supabase);
+  ///private constructor
+  NewsViewModel.create(this.supabase);
```
Thechi2000, line 38: Why does it need to be private ?

lib/models/news.g.dart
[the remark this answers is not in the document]
Pacsort17: I thought I had to push it ^^' So not really, I will delete it in the next push

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+import 'package:path_provider/path_provider.dart';
+
+import 'package:actualia/viewmodels/news.dart';
+
+/// Records news locally to retrieve them offline. Must be created through OfflineRecorder.create(newsProvider)
+class OfflineRecorder {
+  static const ROOT_OFFLINE_FOLDER = "storage";
+
+  late final _appOfflineNewsPath;
+  var _maxStorageSize = 17e+6.toInt(); // One minute mp3 file ~ 1MB
```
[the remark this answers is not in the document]
Pacsort17: It is just random, the idea was if the app has enough space, it should be able to store the last two weeks of news

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+
+    return newStorageSize;
+  }
+
+  int getCurrentMaxStorageSize() {
+    return _maxStorageSize;
+  }
+
+  /// TODO : Only delete some files
+  void _cleanStorage() async {
```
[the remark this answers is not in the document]
Pacsort17: Yes it's what it does, I'll add some documentation

lib/models/offline_recorder.dart
Pacsort17: It is already done, I forgot to delete the todos :/

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+    await file.writeAsString(json);
+
+    // TODO : Only delete oldest files when overflowing
+    if (await _dirSize(Directory(_appOfflineNewsPath)) >= _maxStorageSize) {
+      _cleanStorage();
+    }
+  }
+
+  /// Load data from local storage. Date must be formated in the following way : <year>-<month>-<day>
+  Future<News> loadNews(String date) async {
```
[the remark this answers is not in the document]
Pacsort17: Because I didn't thought about it, I'll update it

lib/models/offline_recorder.dart
Pacsort17: They are used with jsonDecode, but they are not recognized as News (since the list of Paragraphs is not recognized as a list of paragraphs, and that's the way I found to fix that)

lib/viewmodels/news.dart
Pacsort17: Because it has to be created through the factory (init function) due to asynchronous instantiation of the OfflineRecorder

lib/models/offline_recorder.dart
```diff
@@ -0,0 +1,164 @@
...
+    _maxStorageSize = newStorageSize;
+
+    return newStorageSize;
+  }
+
+  int getCurrentMaxStorageSize() {
+    return _maxStorageSize;
+  }
+
+  /// TODO : Only delete some files
```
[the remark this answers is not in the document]
Pacsort17: Also a TODO that I forgot to remove

lib/models/offline_recorder.dart
Pacsort17: To be created asynchronously (needed for creating the folder in in the file system)

Pacsort17:
I made all the changes except the one from Ludovic that says "first fetch from local storage, and then from supabase" since it would mean that local sotrage is more up-to-date than supabase (which is not the case). Is there a way to check the last update of a news in supabase, before requesting the news itself ? ( @pmdlt )

lib/models/offline_recorder.dart
Thechi2000: I'm not sure to understand. I don't see why you couldn't replace this whole chunk of code by `News.fromJson(jsonDecode(json))`. It compiles, but I wasn't able to test it (see the main comment).

lib/models/news.g.dart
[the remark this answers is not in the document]
Thechi2000: Oh in fact it is not generated automatically, so it either need to be in the repo, or be generated automatically in the CI (and explained in the doc). The second option would be best, but you can go for the first if you want.

Thechi2000 (commented):
> I made all the changes except the one from Ludovic that says "first fetch from local storage, and then from supabase" since it would mean that local sotrage is more up-to-date than supabase (which is not the case).

I don't recall saying such a thing. But anyway, it is indeed wrong, the database should be fetched first, or use the local storage if the network is unreachable. If the fetched succeeded, the news should be stored in the local storage.

lib/models/offline_recorder.dart
```diff
@@ -146,25 +147,17 @@ class OfflineRecorder {
...
-
-    String filePath =
-        _appOfflineNewsPath + "${date.year}-$month-${day}_transcript.json";
-    if (!await File(filePath).exists()) {
-      throw FileSystemException("$filePath doesn't exist");
+  Future<News?> _readNewsFile(File file) async {
+    if (!(file.path
+        .substring(file.path.length - "XXXX-XX-XX_transcript.json".length)
+        .startsWith(RegExp(r"[0-9]{4}-[0-9]{2}-[0-9]{2}")))) {
+      return Future(() => null);
```
Thechi2000, line 154:
```suggestion
      return null;
```

Can be simplified since the function is marked async.

lib/models/offline_recorder.dart
```diff
@@ -178,8 +171,44 @@ class OfflineRecorder {
...
+        date.month < 10 ? "0${date.month.toString()}" : date.month.toString();
+    if (date.year < 1000) {
+      throw Exception("The date is too far in the past");
+    }
+
+    String filePath =
+        _appOfflineNewsPath + "${date.year}-$month-${day}_transcript.json";
+    if (!await File(filePath).exists()) {
+      throw FileSystemException("$filePath doesn't exist");
+    }
```
Thechi2000, line 194: I think we could use a builtin date formatter, for example using the ISO norm (like [`toIso8601String`](https://api.flutter.dev/flutter/dart-core/DateTime/toIso8601String.html)). I'm also unsure about the pertinence of the check `year < 1000`.

lib/viewmodels/news.dart
```diff
@@ -116,12 +124,13 @@ class NewsViewModel extends ChangeNotifier {
     } catch (e) {
       try {
         _offlineRecorder
-            .loadNews(date)
-            .then((loadedNews) => _news = loadedNews);
+            .loadAllNews()
+            .then((loadedNews) => setNewsList(loadedNews));
       } catch (e) {
         log("Error fetching news: $e", level: Level.WARNING.value);
```
Thechi2000, line 130: An error should be displayed to the client here, and the log should state that the error occurred during the fetch of cached news (to avoid confusing it with an error fetching the news from the database).

lib/viewmodels/news.dart
```diff
@@ -116,12 +124,13 @@ class NewsViewModel extends ChangeNotifier {
...
-        _news = null;
       }
+      log("Error fetching news: $e", level: Level.WARNING.value);
+      _setError(ErrorType.fetch);
```
Thechi2000, line 133: This error should not be displayed anymore, since it should use the news cached locally. The log may be kept imo, to help debugging.

lib/models/offline_recorder.dart
Thechi2000, line 1: There are a few unused imports in this file.

lib/models/offline_recorder.dart
```diff
@@ -178,8 +171,44 @@ class OfflineRecorder {
...
+    final Iterable<File> files = (await dir.list().toList()).whereType<File>();
+
+    List<News> news = List.empty(growable: true);
+    for (var f in files) {
+      await _readNewsFile(f).then((n) => {
+            if (n != null) {news.add(n)}
+          });
+    }
+
+    return Future.value(news);
```
Thechi2000, line 212:
```suggestion
    return news;
```

Can be simplified as the function is marked async.

Thechi2000 (changes requested):
Good job, especially with the merge on my previous refactor.

However, there are still a few issues, especially with saving the news locally. It is still done upon generating a news manually, which has severe drawbacks (as I said in one of my previous reviews, [here](https://www.github.com/ActualIA/app/pull/86#pullrequestreview-2064184476)).

I've also reopened the thread about using the generated `_NewsFromJson()` function.

lib/models/offline_recorder.dart
Pacsort17: When I replace with News.fromJson(jsonDecode(json)), I have an error that says that it is not a list of paragraphs, and that's the only way that I found to resolve this error :/

lib/models/offline_recorder.dart
Pacsort17: Implemented

lib/viewmodels/news.dart
Pacsort17: Removed

lib/viewmodels/news.dart
Pacsort17: Done

lib/models/offline_recorder.dart
Pacsort17: Yes indeed, thanks !

Thechi2000 (approved):
I've tested both on Android and Web, and the transcripts as well as the audio are correctly stored and can be displayed/played while offline. Good job !

quartztz:
Tested on device, works great !!!! can be merged.

quartztz (approved):
Tested on device, works great !!!! can be merged.

PR #87, opened 2024-04-18
 85 update news view and viewmodel to display all the news of the user

ClementJosso:
This PR changes the news in order to display all of the news (max 10)  a user has in the DB rather than just today's. It should look like this :
image
It would be nice if the reviewer could test it with both an account with multiple news on it, as well as an account which has no news for today, to see if it generates it correctly (these 2 types of account are not incompatible, one can satisfy both criteria).


lib/viewmodels/news.dart
```diff
@@ -56,44 +58,93 @@ class NewsViewModel extends ChangeNotifier {
...
       _news = null;
     }
   }
 
+  Future<void> getNewsList() async {
+    try {
+      var response = await fetchNewsList();
+
+      if (response.isEmpty) {
+        generateAndGetNews();
```
Thechi2000, line 82:
```suggestion
        await generateAndGetNews();
```

lib/viewmodels/news.dart
```diff
@@ -56,44 +58,93 @@ class NewsViewModel extends ChangeNotifier {
...
+        _newsList.insert(0, _news!);
+      } else {
+        _newsList = response.map<News>((news) => parseNews(news)).toList();
+
+        //If the date of the first news is not today, call the cloud function
+        if (_newsList[0].date.substring(0, 10) !=
+            DateTime.now().toString().substring(0, 10)) {
+          generateAndGetNews();
```
Thechi2000, line 90:
```suggestion
          await generateAndGetNews();
```

lib/viewmodels/news.dart
```diff
@@ -56,44 +58,93 @@ class NewsViewModel extends ChangeNotifier {
...
+      log("Error fetching news list: $e");
+      _newsList = [];
+    }
+    notifyListeners();
+  }
+
+  @protected
+  Future<void> generateAndGetNews() async {
+    await invokeTranscriptFunction();
+    await fetchNews(DateTime.now());
```
Thechi2000, line 104: Shouldn't this be `fetchNewsList()`, since we are always dealing with multiple entries now ?

Thechi2000 (changes requested):
When running on Android, I get the following error with my personal account:
```
I/flutter ( 8533): Error fetching news list: type 'Null' is not a subtype of type 'String'
```

On the other hand, the guest account is stuck on the loading after sucessfully generating the transcript. I think this is due to a mistake in the `generateAndGetNews()` function, which only calls `fetchNews()` (see my comment).

lib/viewmodels/news.dart
ClementJosso: In my opinion not necessarily, we could use fetchNewsList(), but this function is actually only called when we need a single entry from the database, so fetchNews() works fine. Maybe in the future we can remove it if we decide to remove the single entry functions altogether, but at the moment I find it useful.

lib/viewmodels/news.dart
```diff
@@ -7,12 +7,19 @@ class NewsViewModel extends ChangeNotifier {
   late final supabase;
   News? _news;
   News? get news => _news;
```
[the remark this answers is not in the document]
ClementJosso: Same here, the parameter news is actually still used because this is where the result of the single entry query is written. We could remove it and make it a return parameter of the query, but I'd rather wait for the offline part of the app to be implemented, because that will probably significantly impact the way this View Model interacts with the DB, so it's good to have everything at hand to know what to keep and what to remove once we have the offline storage.

LeanderCretegny (commented):
I think that the code is good and work as intended, but I think that we should change the title for the news from previous days.

LeanderCretegny (approved):

PR #95, opened 2024-04-20
94 feat add multiple providers in generate news function

### LINKED ISSUE #94: feat: add multiple providers in generate news function

Thechi2000:
Closes #94 

Adds multiple provider management in Supabase, by using a dedicated type (`Provider`). It is stored in the database as a JSON. Also adds a `fetchNews()` function, to fetch normalized news from any provider, to ease transcript generation.


test/unit/providers.dart
pmdlt, line 1: Not sure if we need to commit this one, as I'm planning to clean up the Supabase folder, and we've agreed on the testing issue. If we leave a few tests, it's a mess, isn't it?

pmdlt (dismissed):
As for the rest, everything seems perfect! Thanks for the job as always :)
I'm going to ask the rest of the group to review this one quickly, so I can get on with reformatting it all.

test/unit/providers.dart
Thechi2000: I'm not sure to understand. Those tests only check that the providers are correctly parsed, to ease the integration with the backend. I added them since I defined the format of the providers' data in this branch.

lib/models/providers.dart
```diff
@@ -0,0 +1,30 @@
...
+class GNewsProvider extends NewsProvider {}
+
+class RSSFeedProvider extends NewsProvider {
+  final String url;
+
+  RSSFeedProvider({required this.url});
+}
+
+abstract class NewsProvider {
+  static NewsProvider? parse(dynamic dict) {
```
LeanderCretegny, line 13: Suggestion: deserialize would be a better name

LeanderCretegny (commented):
Overall great, minor addition suggested

LeanderCretegny (approved):

quartztz (dismissed):
Looks great ! Code style matches guidelines, as well. Approved!

test/unit/providers.dart
pmdlt: Let's leave it at that then!

pmdlt:
image

I'm getting weird errors using user `48ecf965-4688-4b19-8af1-3df2fa50282b` though. I'll look tomorrow to see if I can get more info on the cause of this error.

pmdlt:
> image
> I'm getting weird errors using user `48ecf965-4688-4b19-8af1-3df2fa50282b` though. I'll look tomorrow to see if I can get more info on the cause of this error.

Update on that, it seems to be working well after all. I'm not in a position to reproduce this error on your branch.
It was certainly due to my previous modifications

pmdlt:
image
[supabase_logs.csv](https://github.com/ActualIA/app/files/15132598/supabase_logs.1.csv)

This doesn't seem to work in deployment (see image and logs), no news is returned even though the user used has several news providers.

Thechi2000:
It seems like the `news_provider` table was reverted to an old version, therefore with an incorrect schema and no rows. After fixing it and adding back the provider, the edge function is working correctly.

However, the edge function is currently failing if the user has undefined or no providers. Perhaps we would like to default to Google News in that case, what do you think ?

pmdlt:
> It seems like the `news_provider` table was reverted to an old version, therefore with an incorrect schema and no rows. After fixing it and adding back the provider, the edge function is working correctly.
> 
> However, the edge function is currently failing if the user has undefined or no providers. Perhaps we would like to default to Google News in that case, what do you think ?

Yes indeed, I think we can fetch Google News on general topics like politics, science and technology.
I'm now testing it on my side to see if it works again.

pmdlt:
When I test on the same user, I get :

```
Log ID
ad6f6947-0a69-4252-9a77-0e7827b65764

Log Timestamp (UTC)
2024-04-27T12:22:56.430Z

Log Event Message
TypeError: Cannot read properties of undefined (reading 'title')
    at file:///home/deno/functions/get-transcript/index.ts:89:22
    at Array.forEach (<anonymous>)
    at generateArticlesPrompt (file:///home/deno/functions/get-transcript/index.ts:88:10)
    at generateTranscript (file:///home/deno/functions/get-transcript/index.ts:123:20)
    at Object.handler (file:///home/deno/functions/get-transcript/index.ts:59:46)
    at eventLoopTick (ext:core/01_core.js:64:7)
    at async handleHttp (ext:sb_core_main_js/js/http.js:112:15)
```
Is this related to what you said above? This user has no providers?
Not according to the rest of the logs.

Thechi2000:
I was not able to reproduce the bug you encountered, even with the same user (and query parameters). I've set the default providers to GNews and topics to politics and technology.

pmdlt (approved):
As seen together, it looks like the cause was external to the code. All's well on my side then, after several successive tests

Pacsort17 (approved):
The script in supabase/functions/providers.ts at lines127 to 129 should be a little more complete to avoid having several times the same topic (and maybe some prefered topics), but it is another issue.

Tested on android (and edge), but the views are not functionning (news loading error + when clicking on the news the screns becomes red and displays "RangeError (end): Invalid value: Only valid value is 0: 10". However, I don't think it comes from this pr but adding tests on this view when there is an error has to be done (in another branch) (and implementing the fat that clicking on the "Something went wrong" message does nothing instead of returning an error).

Tests seems to be complete, nice job !

PR #96, opened 2024-04-21
chore: fix format check hook

Thechi2000:
Fix the client hook to correctly detect non-formatted code. Also fix color persistence in the shell after a failure.

### DIFF
--- a/dart-format-check
+++ b/dart-format-check
@@ -1,14 +1,14 @@
 #!python
 
-import subprocess
 import sys
+import os
 
 if len(sys.argv) == 1:
     exit(0)
 
-ret = subprocess.run(["dart", "format", "-o", "none", "--set-exit-if-changed", *sys.argv[1:]], shell=True)
+ret = os.system("dart format --set-exit-if-changed -o none " + " ".join(sys.argv[1:]))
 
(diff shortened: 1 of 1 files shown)


pmdlt:
@Thechi2000 Why not directly `dart format lib` ?
And prompt an error if not fixed (witch is a possibility)

Thechi2000:
> @Thechi2000 Why not directly `dart format lib` ? And prompt an error if not fixed (witch is a possibility)

I personally don't like having hooks modifying the commit, even if it says it. I prefer knowing exactly what is committed, without having to check afterwards. But I understand, and we can do it if the majority wants it.

Thechi2000:
> @Thechi2000 Why not directly `dart format lib` ? And prompt an error if not fixed (witch is a possibility)

I personally don't like having hooks modifying the commit, even if it says it. I prefer knowing exactly what is committed, without having to check afterwards. But I understand, and we can do it if the majority wants it.

pmdlt (approved):
Okay :) so go on, lgtm

ClementJosso (approved):
Tested on Windows, commits with incorrect code format are indeed blocked and the red color of the warning doesn't stay after :)

PR #100, opened 2024-04-24
99 feat implement backend for alarm management

pmdlt:
**Introducing Periodic Transcript and Audio Generation via Alarms**

This pull request introduces a new functionality aimed at automating the generation of transcripts and corresponding audio files.

**Purpose**:

The primary goal of this feature is to alleviate manual intervention in the transcript and audio generation process, ensuring timely updates and availability of these essential assets. By periodically invoking the `generate-via-alarms` function, transcripts and audio files are generated.

**Key Additions**:

1. **Edge Function**: 
    - Implemented a new Supabase Edge function, `generate-via-alarms`, to orchestrate the generation process.
    - Will be called every hours/30min.

2. **Database Integration**:
image

    

**Integration Requirements**:

To seamlessly integrate this functionality into the Flutter app, you just need to update the DB with required fields (see above).

**Instructions for Testing Locally**:

To test this functionality locally, follow these steps:

1. Run `supabase start`  and `supabase functions serve` to start the Supabase server locally.
2. Make an HTTP POST request to the `generate-via-alarms` function endpoint with appropriate headers and payload.


pmdlt:
⚠️ This PR need the presence of `generate-transcript` updated version

supabase/functions/generate-via-alarms/index.ts
```diff
@@ -74,33 +76,21 @@ Deno.serve(async (req) => {
...
 
-    if (generateTranscriptResponse.error) {
-      console.error(
-        "Error generating transcript:",
-        generateTranscriptResponse.error.message,
-      );
-      return new Response("Error generating transcript", { status: 500 });
-    }
-
     console.log(generateTranscriptResponse);
```
Thechi2000, line 84: Is it really required to log the full transcript ? We could simply log the success/failure of the call IMO. Also, we have to log the errors, and with the correct level.

supabase/functions/generate-via-alarms/index.ts
```diff
@@ -74,33 +76,21 @@ Deno.serve(async (req) => {
...
-      );
-      return new Response("Error generating audio", { status: 500 });
-    }
+    const generateAudioResponse = await generateAudio(
+      transcriptId,
+      "echo",
+      supabaseClient,
+    );
 
     console.log(generateAudioResponse);
```
Thechi2000, line 95: See above.

Thechi2000 (commented):
Like we discussed in #124, shouldn't the Edge Functions use UTC time instead of local one ? Furthermore, local time is the time of the Supabase instance, not the one of the actual user.

Otherwise, good job ! I feel like the modularity added will be useful.

quartztz (approved):
Nice ! Unblocks a bunch of work about alarms. Can be merged !

supabase/functions/generate-via-alarms/index.ts
pmdlt: Sure, I'm going to delete these logs. The rest is already logged in each function separately.

pmdlt: Sure, I'm going to delete these logs. The rest is already logged in each function separately.

quartztz (approved):
Agree with @Thechi2000's comments. LGTM!

Thechi2000 (approved):

PR #108, opened 2024-04-25
106 Implement audio playing for daily news

ClementJosso:


ClementJosso (commented):
This commit is only meant as a way to let Jacopo work on it with me, it doesn't add a completed feature

ClementJosso:
As of right now, we are making progress towards having a working audio player. We can know play files that we write on the device, but the pause doesn't work, and the call to the generate-audio function doesn't work as well

lib/widgets/play_button.dart
```diff
@@ -0,0 +1,88 @@
...
+        icon: _playerState == PlayerState.playing
+            ? const Icon(
+                Icons.pause_circle_outline,
+                size: 40.0,
+                color: Color.fromARGB(255, 68, 159, 166),
+              )
+            : const Icon(
+                Icons.play_circle_outline,
+                size: 40.0,
+                color: Color.fromARGB(255, 68, 159, 166),
```
Thechi2000, line 53: Couldn't we use colors from the App's theme, for consistency ?

test/widgets/play_button.dart
```diff
@@ -0,0 +1,75 @@
...
+            home: _child));
+  }
+}
+
+void main() {
+  // The `BuildContext` does not include the provider
+  // needed by Provider<AuthModel>, UI will test more specific parts
+  testWidgets('testPlayButton', (WidgetTester tester) async {
+    // Build our app and trigger a frame.
+    // await tester.pumpWidget(const App());
```
Thechi2000, line 47: Those comments seems out of place, probably a copy-paste ?

lib/viewmodels/news.dart
```diff
@@ -161,12 +170,76 @@ class NewsViewModel extends ChangeNotifier {
     }
   }
 
+  // Function to get the audio file from the database
+  Future<void> getAudioFile(News news) async {
+    var audio = news.audio;
+    if (news.transcriptId == -1) {
```
pmdlt, line 176: I think it would be worth adding comments here, I didn't understand at first what this check did.

test/widgets/source_view.dart
```diff
@@ -27,7 +27,7 @@ void main() {
         "facilisis sem. In massa urna, rutrum eu turpis a";
     const String title = "lorem ipsum dolor sit amet, consectetur";
     const String newsPaper = "https://dummy.com";
-    const String date = "00/00/0000";
+    const String date = "1111-11-11";
```
Thechi2000, line 30: We should probably use dates after the epoch (01/01/1970) to avoid potential issuesn

lib/viewmodels/news.dart
```diff
@@ -161,12 +170,76 @@ class NewsViewModel extends ChangeNotifier {
...
+      return;
+    }
+    audio ??= await generateAudio(news.transcriptId);
```
pmdlt, line 179: How does error handling work here? What happens if the Edge generate-audio function returns something other than 200?

lib/viewmodels/news.dart
```diff
@@ -161,12 +170,76 @@ class NewsViewModel extends ChangeNotifier {
...
+    final filePath = '${directory.path}/audios/$transcriptId.mp3';
+
+    final file = File(filePath);
+    if (await file.exists()) {
+      return DeviceFileSource(filePath);
+    } else {
+      log("Can't find audio file at $filePath", level: Level.WARNING.value);
+      return null;
+    }
+  }
```
Thechi2000, line 235: Those lines are not covered, isn't there a way to do this ? Perhaps by using dependency injection for `getApplicationDocumentsDirectory()`.

lib/viewmodels/news.dart
pmdlt: I mean, wouldn't it be worth putting the await in the try, and displaying a clear error message to the user?

pmdlt: (you can also decide to leave it for another PR)

lib/viewmodels/news.dart
```diff
@@ -161,12 +170,76 @@ class NewsViewModel extends ChangeNotifier {
...
+
+      if (!await transcriptsDirectory.exists()) {
+        await transcriptsDirectory.create(recursive: true);
+      }
+
+      final file =
+          File('${transcriptsDirectory.path}/${news.transcriptId}.mp3');
+      await file.writeAsBytes(response);
+    } catch (e) {
+      log('Error downloading audio file: $e', level: Level.WARNING.value);
```
pmdlt, line 202: do you think it's possible to display a clear error to the user?

Thechi2000 (commented):
Works great overall ! The automatic download of the audio files is great to avoid latency, but I think we could have a toast to tell the user when it is not already available.

lib/viewmodels/news.dart
ClementJosso: Do you mean displaying it on the screen ? We could but I don't see any case where this error could happen to be honest, we're catching the error in case it happens but since we corrected the function it has never thrown any

lib/viewmodels/news.dart
ClementJosso: What do you mean other than 200 ? Is it a code the function sends on success ?

supabase/functions/generate-transcript/index.ts
```diff
@@ -25,22 +31,37 @@ interface Transcript {
...
+  try {
+    const body = await request.json();
+    const [passes, errors] = await validate(body, bodySchema);
+    if (!passes) {
+      console.log(`Invalid body: ${JSON.stringify(errors)}`);
+      return new Response(JSON.stringify({ errors }), { status: 400 });
+    }
+
+    userId = body.userId;
+  } catch (_) {}
```
pmdlt, line 56: Why a try/catch here if we don't care of the catch ? @Thechi2000

pmdlt (changes requested):

lib/viewmodels/news.dart
pmdlt: Yes, 200 on success, 400, 500, etc,. on errors

lib/viewmodels/news.dart
pmdlt: An Internet problem, for example. But yes, @Thechi2000  is right, you can do that in a specific PR that takes care of displaying nice errors

supabase/functions/generate-transcript/index.ts
ClementJosso: I suggest we do theses changes in another PR since it is not really related to the objective of this branch

ClementJosso:
I took notes of your comments and made changes. However, the toasts will be added in another PR, and I am unable to make some of the tests requested at the moment.

pmdlt (approved):
Thanks for the changes !

Thechi2000 (dismissed):
Let's merge that !

Thechi2000 (approved):

loggerz (approved):

PR #113, opened 2024-04-26
111 refactor: handle onboarding status in the auth state listener

### LINKED ISSUE #91: Login persistence
Login is not persisted when the user closes the App, requiring to repeat the authentication workflow each time the app is opened.

### LINKED ISSUE #111: refactor: store onboarding status in supabase user metadata
Instead of using a computed value based on the presence of a NewsSettings for the current user, we should store the status using the [metadata feature of Supabase](https://supabase.com/docs/guides/auth/managing-user-data?queryGroups=language&language=kotlin#accessing-user-metadata)

loggerz:
This small refactor  will help the extension of the onboarding views by moving the "onboarding required" flag to the authentication model (fixes #111) .

The simpler handling of the authentication also fixes #91 by simply re-loading the restored user from the Supabase SDK

**Instructions for Testing Locally:**
- Onboarding status stored in metadata
  - Logout
  - Go to the [authentication page on Supabase console](https://supabase.com/dashboard/project/dpxddbjyjdscvuhwutwu/auth/users) and delete your user
  - Log back in, the onboarding is expected to show up
  - Logout, log back in, the news should already be displayed (and not the onboarding)
- Login persistence
  - Login
  - Kill the app with your task manager to force a cold start
  - The news should already be displayed (and not the login page)

### DIFF
--- a/lib/models/auth_model.dart
+++ b/lib/models/auth_model.dart
@@ -87,11 +87,7 @@ class AuthModel extends ChangeNotifier {
   Future<bool> signInWithEmail(String email, String password) async {
     final res = await _supabase.auth
         .signInWithPassword(password: password, email: email);
-    if (res.user != null) {
-      // isOnboardingRequired = true; // always show onboarding with test (non-google) accounts
-      return true;
-    }
-    return false;
+    return res.user != null;
   }
 
   Future<bool> signInWithFakeAccount() async {
--- a/lib/models/auth_model.dart
+++ b/lib/models/auth_model.dart
@@ -12,16 +12,20 @@ class AuthModel extends ChangeNotifier {
   StreamSubscription? authStateSub;
 
   bool get isSignedIn => user != null;
+  late bool isOnboardingRequired;
 
   AuthModel(SupabaseClient supabaseClient, this._googleSignIn) {
     _supabase = supabaseClient;
     authStateSub = _supabase.auth.onAuthStateChange.listen((authState) {
       print('Supabase onAuthStateChange new authState : ${authState.event}');
 
-      if (authState.event == AuthChangeEvent.signedIn) {
-        user = authState.session!.user;
-      } else if (authState.event == AuthChangeEvent.signedOut) {
-        user = null;
+      user = authState.session?.user;
+
+      if (user != null) {
+        print(user!.userMetadata);
+        final bool onboardingDone =
+            user!.userMetadata?['onboardingDone'] ?? false;
+        isOnboardingRequired = !onboardingDone;
       }
 
       notifyListeners();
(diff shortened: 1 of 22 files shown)


lib/models/auth_model.dart
```diff
@@ -12,16 +12,25 @@ class AuthModel extends ChangeNotifier {
...
-      } else if (authState.event == AuthChangeEvent.signedOut) {
+      if (authState.event == AuthChangeEvent.signedOut ||
+          authState.event == AuthChangeEvent.userDeleted) {
         user = null;
+      } else {
+        user = _supabase.auth?.currentSession?.user;
+      }
+
+      if (user != null) {
+        print(user!.userMetadata);
```
Thechi2000, line 30:
```suggestion
```

Should be removed, since it contains confidential information.

lib/models/auth_model.dart
```diff
@@ -79,10 +88,21 @@ class AuthModel extends ChangeNotifier {
   Future<bool> signInWithEmail(String email, String password) async {
     final res = await _supabase.auth
         .signInWithPassword(password: password, email: email);
-    return res.user != null;
+    if (res.user != null) {
+      // isOnboardingRequired = true; // always show onboarding with test (non-google) accounts
```
Thechi2000, line 92: Why is this commented ? I think the line should either be uncommented (always show onboarding), or completely removed (only on first time).

lib/models/auth_model.dart
```diff
@@ -79,10 +88,21 @@ class AuthModel extends ChangeNotifier {
...
+    }
+    return false;
   }
 
   Future<bool> signInWithFakeAccount() async {
     return signInWithEmail("actualia@example.com", "actualia");
   }
+
+  /// Sets a flag in user metadata to remember that the onboarding is done
+  Future<bool> setOnboardingIsDone() async {
```
Thechi2000, line 103: I find the name a bit weird. I feel like a setter should take a value parameter, so we would have `setOnboardingCompletion(boolean completed)`, or `markOnboardingAsDone()` to only set it to true.

Thechi2000 (changes requested):
I tested on Android, everything works as expected. Good job !

I just have a few small comments on the code.

lib/models/auth_model.dart
LeanderCretegny: If the onboarding is executed each time it could be annoying depending on what you are implementing or testing, but you also might need it to always be executed, hence the commented line that can simply be  uncommented.

lib/models/auth_model.dart
LeanderCretegny: As you said, we don't want to be able to mark it as anything else than true, hence the parameter is useless. Then regarding your renaming proposition, I find it completely equivalent.

lib/models/auth_model.dart
Thechi2000: IIUC, the line is only commented for debugging purpose. If this si correct, then it should be uncommented in the production code.

loggerz: It was not for debugging purprose, it was an older implementation. It should be removed, we might want to override the behavior for test account as suggested by the comment but this hasn't been clearly decided yet.
Thanks for catching this

Thechi2000 (approved):

quartztz (approved):
Had the same comments as Ludo before changes were made, glad to see it's fixed now :D looks good to me!

PR #117, opened 2024-04-29
104 feat onboarding provider update

LeanderCretegny:
This feature integrate the news providers selection to the onboarding wizard and to the settings.

**To test**: 
 - Go to supabase and delete your google account to simulate a first  connection to the app or set isOnboardingRequired to true in auth_model.dart, line 34.
 - Launch the app and connect yourself with your google account
 - Complete the onboarding wizard
 - Once wizard is completed you should be on the news page 
 - then go to the settings, and click on Sources button, check that you can see the providers you entered during onboarding.


lib/models/providers.dart
```diff
@@ -43,7 +43,7 @@ abstract class NewsProvider {
   /// Converts a dynamic value to a NewsProvider.
   ///
   /// The following format are accepted:
-  /// - GNews: `{"type", "gnews"}`
+  /// - GNews: `{"type", "gnews"}` todo change ',' by ':'
```
Thechi2000, line 46:
```suggestion
  /// - GNews: `{"type": "gnews"}`
```

lib/widgets/wizard_widgets.dart
```diff
@@ -37,14 +39,11 @@ class _WizardSelector extends State<WizardSelector> {
...
-    Widget title = Text(
-        style: Theme.of(context).textTheme.titleLarge,
-        textAlign: TextAlign.center,
-        widget.title);
+    Widget title = WizardSelectorTitle(title: widget.title);
 
     Widget body = Expanded(
         child: Container(
-            padding: const EdgeInsets.symmetric(vertical: UNIT_PADDING * 2),
+            padding: const EdgeInsets.fromLTRB(0.0, 24.0, 0.0, 24.0),
```
Thechi2000, line 46: Should you the theme's values, not hardcoded ones.

lib/widgets/wizard_widgets.dart
```diff
@@ -79,6 +78,45 @@ class _WizardSelector extends State<WizardSelector> {
...
+    return Container(
+      margin: const EdgeInsets.fromLTRB(16, 50, 16, 8.0),
+      child: Text(
+        text,
+        textScaler: const TextScaler.linear(2.0),
+        style: const TextStyle(
+            fontFamily: 'EB Garamond',
+            fontWeight: FontWeight.w700,
+            color: Colors.black),
+        maxLines: 2,
```
Thechi2000, line 97: Should you the theme's values, not hardcoded ones.

lib/widgets/wizard_widgets.dart
```diff
@@ -79,6 +78,45 @@ class _WizardSelector extends State<WizardSelector> {
...
+
+  const WizardSelectorTitle({this.title = "default title", super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Text(
+        style: const TextStyle(
+            fontFamily: "Fira Code",
+            fontWeight: FontWeight.w700,
+            fontSize: 32.0),
```
Thechi2000, line 114: Should you the theme's values, not hardcoded ones.

lib/widgets/wizard_widgets.dart
```diff
@@ -79,6 +78,45 @@ class _WizardSelector extends State<WizardSelector> {
...
+        maxLines: 2,
+      ),
+    );
+  }
+}
+
+class WizardSelectorTitle extends StatelessWidget {
+  final String title;
+
+  const WizardSelectorTitle({this.title = "default title", super.key});
```
Thechi2000, line 106: Does it really need a default title ? If yes, it would be more adapted to have an actual title instead of a placeholder.

lib/widgets/wizard_widgets.dart
```diff
@@ -111,3 +149,139 @@ class WizardNavigationBottomBar extends StatelessWidget {
...
   }
 }
+
+class WizardScaffold extends StatelessWidget {
+  final PreferredSizeWidget topBar;
+  final Widget body;
+
+  const WizardScaffold(
+      {this.topBar = const TopAppBar(),
+      this.body = const Text("unimplemented"),
```
Thechi2000, line 159: Like above, does it really need a default value ?

lib/widgets/wizard_widgets.dart
```diff
@@ -111,3 +149,139 @@ class WizardNavigationBottomBar extends StatelessWidget {
...
+      super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+        appBar: topBar,
+        body: Container(
+          padding: const EdgeInsets.fromLTRB(48.0, 48.0, 48.0, 48.0),
```
Thechi2000, line 167: Should you the theme's values, not hardcoded ones.

lib/widgets/wizard_widgets.dart
```diff
@@ -111,3 +149,139 @@ class WizardNavigationBottomBar extends StatelessWidget {
...
+      ],
+    );
+
+    Widget displaySelected = Expanded(
+        child: Container(
+      padding: const EdgeInsets.fromLTRB(0.0, 24.0, 0.0, 24.0),
+      child: SingleChildScrollView(
+        child: Wrap(
+          spacing: 8.0,
+          runSpacing: 8.0,
```
Thechi2000, line 251: Should you the theme's values, not hardcoded ones.

Thechi2000 (changes requested):
I did spot the following bugs:
- When going through the source's settings, the entries are added each time, without removing the old ones. When logging out then back in, I have to click "Google News several times to get rid of it.

On a clean user, I added Google News and `https://test.rss`, then went through the settings 2 times (without changing anything) and got:
image

Also, it is not very clear that the user has to type "Enter" to validate a RSS Feed, maybe something could be improved ? It took me to attempts where I thought it just did not work.

Also, I don't really like the way you handle the different selectors. I find it very error prone and dirty. Instead, I think that each selector should be associated with a metadata - the provider - and the display should directly use the `displayString()` method. This way, there would be no need of serializing and then deserializing the values. The code would be much nicer that way IMO.

lib/widgets/wizard_widgets.dart
LeanderCretegny: In this case I find it relevant to act as a placeholder to help during development and testing. It is true that it is not required but it does not hurt and could even prove useful :)

Thechi2000 (dismissed):
Thanks for the changes, the code is much cleaner that way ! All the bugs are fixed now, good work.

quartztz (approved):
looks great on my end!

Thechi2000 (approved):

pmdlt (approved):

PR #118, opened 2024-04-30
115 test end to end tests for the app

### LINKED ISSUE #115: test: end to end tests for the app
We need at least one end to end test for the app. Each one should be a full walkthrough for one feature of the application:
- [x] Onboarding
- [x] Settings edition
- [x] Source consultation

Thechi2000:
Adding the first end to end test, for the onboarding process. It is run along with the other tests when using the `test.sh` script, and can be run independently while display the UI by using `flutter run test/integration/onboarding.dart`.

Since flutter blocks all HTTP requests when testing, it uses a mocked HTTP client to simulate the requests to Supabase.

Closes #115


Pacsort17 (dismissed):
Thanks for the print cleaning :') 
Test passes locally on Windows.

In the future, the end to end test should at least test if a transcript is generated, and if the source is correct, even if it means emulation a supabase client.

Small refactor : "dpxddbjyjdscvuhwutwu" should be stored in a variable instead of copy pasting it, so it only has to be changed once if needed.

Thechi2000:
> Small refactor : "dpxddbjyjdscvuhwutwu" should be stored in a variable instead of copy pasting it, so it only has to be changed once if needed.

Good idea ! I've extended it to many magic values, like the test user's uuid, the user's data and its access token.

Pacsort17 (dismissed):
Nice refactor :)

Pacsort17 (approved):
Thanks for the additionnal tests !

ClementJosso (approved):
Great tests, I see that they all pass on the CI, so this will be a significant improvement in the coverage. Also, thanks for removing the prints

quartztz (approved):
This is monumental work ! thank you so much, looks amazing

PR #123, opened 2024-05-01
123 handle empty transcripts in news generation

ClementJosso:
Small PR to fix a problem when the DB contains empty transcripts

### DIFF
--- a/lib/viewmodels/news.dart
+++ b/lib/viewmodels/news.dart
@@ -122,14 +122,18 @@ class NewsViewModel extends ChangeNotifier {
 
   News parseNews(dynamic response) {
     List<dynamic> newsItems = response['transcript']['news'];
-    List<Paragraph> paragraphs = newsItems.map((item) {
-      return Paragraph(
-          transcript: item['transcript'],
-          source: item['source']['name'],
-          title: item['title'],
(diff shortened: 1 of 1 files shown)


pmdlt (dismissed):

lib/viewmodels/news.dart
```diff
@@ -122,14 +122,18 @@ class NewsViewModel extends ChangeNotifier {
...
+    for (dynamic item in newsItems) {
+      if (item['transcript'] != null) {
+        paragraphs.add(Paragraph(
+            transcript: item['transcript'],
+            source: item['source']['name'],
+            title: item['title'],
+            date: item['publishedAt'],
+            content: item['content']));
+      }
+    }
```
Thechi2000, line 136: Couldn't we keep a functional approach, e.g. by using `where()` then `map()` ? I would find it nicer that way.

pmdlt (approved):

lib/viewmodels/news.dart
ClementJosso: Good idea, I corrected the code accordingly

loggerz (approved):
That's way cleaner, nice catch Ludo

PR #124, opened 2024-05-01
fix: convert DateTime to correct timezone

Thechi2000:
Convert the dates used by Supabase to the correct timezone. Supabase uses UTC, so each date given or taken from it needs to be converted from or to the local timezone.

### DIFF
--- a/lib/viewmodels/news.dart
+++ b/lib/viewmodels/news.dart
@@ -47,6 +47,9 @@ class NewsViewModel extends ChangeNotifier {
 
   /// Fetches news for the specified date from the database.
   Future<void> fetchNews(DateTime date) async {
+    // Converts time to UTC to match Supabase's instances.
+    date = date.toUtc();
+
     String dayStart =
         DateTime(date.year, date.month, date.day).toIso8601String();
     String nextDayStart =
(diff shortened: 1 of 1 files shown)


quartztz (approved):
It seems like a good short term fix for now, I'll admit. How much can this mess with the alarm generation that's being implemented soon if not right now ? I wouldn't want us to have to rethink how we do this fully due to bugs later.

Thechi2000:
> It seems like a good short term fix for now

I don't think it is anything of a short term fix. Handling multiple timezone is a clear issue for our project, and I think that normalizing them when used in Supabase (which is the same for everyone) is necessary.

We will need to take special care to consistently do it on the client, and use the UTC time on the server, since it may be distributed around the world. In my opinion, this is the easiest way to do it, if we do not want to store the timezone along with the timestamps (which would be a mess).

lib/viewmodels/news.dart
```diff
@@ -87,7 +91,7 @@ class NewsViewModel extends ChangeNotifier {
 
         //If the date of the first news is not today, call the cloud function
         if (_newsList[0].date.substring(0, 10) !=
```
pmdlt, line 93: Not much to do with this PR, but as I see this code in particular I'll take advantage of it: in the future we'll have to change the way we decide to retrieve news. I'd prefer to use a calculation like “if there's been no news for 12 hours, etc.”.

pmdlt (approved):
Thanks for the fix :)

I'll take this opportunity to clarify @quartztz's comment: the alarm system is also based on timeszones, but this time for hours (timetz instead of timestamptz). And this PR does not affect the alarm.
But @loggerz and @ClementJosso will have to remember to deal with this when declaring alarms in Supabase. As far as I'm concerned, Edge function support is already functional.

ClementJosso (approved):
Upon testing, I can see it still works as before during the day, but I can't see for sure if it fixes the problems we had, since it only happened between 12pm and 2am.

PR #131, opened 2024-05-03
126 feat improve news view ux

ClementJosso:
In this PR, we add 3 features to improve the user's experience. First, upon creation of an account, instead of generating a news instantly, we display a view with a button to generate the transcript if wanted, or we can just wait for the alarm.
Second feature, the paragraphs in the news view now have a small text to tell the user they can see the source. 
The third feature, the source view now has a button to go to the webpage of the original article.
It also makes small changes in the news view model to improve functionality.


lib/views/no_news_view.dart
```diff
@@ -0,0 +1,74 @@
...
+                Text(
+                  widget.title,
+                  style: Theme.of(context).textTheme.titleLarge,
+                  textAlign: TextAlign.center,
+                ),
+                const SizedBox(height: UNIT_PADDING * 2),
+                subtitle,
+              ],
+            )),
+        button,
```
LeanderCretegny, line 70: As subtitle and button widget are defined outside of the return statement, I would suggest to also define the return widget outside, for more coherent code.

lib/views/source_view.dart
```diff
@@ -58,9 +63,33 @@ class _SourceViewState extends State<SourceView> {
...
   }
+
+  Future<void> _launchInBrowser(Uri url) async {
+    if (!await launchUrl(
+      url,
+      mode: LaunchMode.externalApplication,
+    )) {
+      throw Exception('Could not launch $url');
+    }
+  }
```
LeanderCretegny, line 94: There should not be any declared function in a State<Widget> object.

pubspec.lock
LeanderCretegny, line 1: I think we should not push this file

pubspec.yaml
LeanderCretegny, line 1: I think we should not push this file

LeanderCretegny (changes requested):
Overall good except for the function declared in the State object, I think this is a bad practice

lib/views/no_news_view.dart
ClementJosso: Good point, I'll adjust accordingly

lib/views/source_view.dart
ClementJosso: I believe this is fine, because the example given for the url_launcher package uses a help function inside the state of their widget as well. See https://pub.dev/packages/url_launcher/example

pubspec.yaml
loggerz: I don't understand what you meant by that as both pubspec files must always be pushed

loggerz (approved):
Fully tested on android device, everything is working as expected

PR #137, opened 2024-05-06
Refactor provider management

### LINKED ISSUE #135: feat: Telegram provider
Add Telegram as a news provider. The app should be able to use messages sent through telegram channels to be used as news for the transcript generation.

DoD: The user can add/select channels used as a news provider for the transcript.

Thechi2000:
Following the discovery of [RSSHub](https://docs.rsshub.app/) by @pmdlt, I used it to refactor the management of the providers.

Now, each provider can be uniquely identified by either a full URL (for a RSS Feed) or a URI on RSS Hub. This URI may contain parameters.

On the database, this changes makes the `news_settings.providers_id` column and the `news_providers` table useless, since everything can be stored in a single `news_settings.providers` text array.

Closes #135

In the future, it could be extended to handle more providers, improve ease of use (e.g. use a telegram invite link) and validate user input.


lib/views/providers_wizard_view.dart
```diff
@@ -1,114 +1,94 @@
...
+              onPressed: () async {
+                pvm.setNewsProviders(items.map((e) => e.toProvider()).toList());
+                if (!await pvm.pushNewsProviders() && Platform.isAndroid) {
+                  Fluttertoast.showToast(msg: "Error while updating providers");
+                } else if (context.mounted) {
+                  Navigator.pop(context);
+                }
+              },
+              child: Text(widget.hasNext ? "Next" : "Done"))
+        ]),
```
LeanderCretegny, line 88: Why not using widget WizardNavigationBottomBar from wizard_widgets.dart?

lib/views/providers_wizard_view.dart
```diff
@@ -1,114 +1,94 @@
...
-              await auth.setOnboardingIsDone();
-            }
-          } catch (e) {
-            log("error in news provider wizard: $e");
-          }
-          Navigator.pop(context);
-        });
+    Widget body = const LoadingView(text: "Loading your sources");
+    if (widget.items != null) {
+      var items = widget.items as List<ProviderWidget>;
```
LeanderCretegny, line 49: Why creating a local variable that will then be modified using setState()? It should be a private attribute of the State<> object holding the state for the item list mentioned in previous comment.

lib/widgets/wizard_widgets.dart
```diff
@@ -171,116 +171,75 @@ class WizardScaffold extends StatelessWidget {
...
-  State<RSSSelector> createState() => _RSSSelector();
-}
-
-class _RSSSelector extends State<RSSSelector> {
-  late TextEditingController _controller;
-  List<(NewsProvider, String)> _selectedItems = [];
+class ProviderWidget extends StatefulWidget {
+  ProviderType type;
+  late List<String> values;
+  void Function(ProviderWidget) onDelete;
```
LeanderCretegny, line 177: attributes should be final (warning raised by my IDE: This class (or a class that this class inherits from) is marked as 'immutable', but one or more of its instance fields aren't final)

lib/widgets/wizard_widgets.dart
```diff
@@ -171,116 +171,75 @@ class WizardScaffold extends StatelessWidget {
...
-    );
+    var fields = widget.type.parameters.map((e) {
+      var index = widget.type.parameters.indexOf(e);
+      return TextField(
+          style: Theme.of(context).textTheme.bodyMedium,
+          decoration: InputDecoration(hintText: e),
+          autocorrect: false,
+          maxLines: 1,
+          controller: TextEditingController(text: widget.values[index]),
+          onChanged: (v) => widget.values[index] = v);
```
LeanderCretegny, line 209: Seems to have forgotten about setState here, and same point as above, I don't like the idea of modifying widget attribute using setState(), imo we should use a local variable inside the State<> object.

lib/widgets/wizard_widgets.dart
```diff
@@ -171,116 +171,75 @@ class WizardScaffold extends StatelessWidget {
...
+        items: ProviderType.values
+            .map((e) => DropdownMenuItem(
+                  value: e,
+                  child: Text(e.displayName),
+                ))
+            .toList(),
+        onChanged: (e) => setState(() {
+              widget.type = e!;
+              widget.values = List.filled(widget.type.parameters.length, "");
+            }));
```
LeanderCretegny, line 223: same point as above, I don't like the idea of modifying widget attribute using setState(), imo we should use a local variable inside the State<> object.

LeanderCretegny (changes requested):

Thechi2000:
All done, thanks for the tips ! I delegated handling the state of the edition to the ViewModel (as it should have been).

LeanderCretegny (changes requested):
I have encountered a bug when running the code on my emulator, the provider widget is overflowing

Thechi2000:
Thanks for the catch ! I fixed it, as well as replaced an `ListView` by a `Column` in the provider's widget.

LeanderCretegny (approved):
All bug found corrected, and code is great!

pmdlt:
I have several exceptions that happen when launching the app from the web and manipulating this view (creating and deleting sources), but I think it's related to the platform.

image

pmdlt (approved):
Update: works well on an APK build generated for the occasion.
Ideally, we'd like to be able to choose Google News (or other providers that can only be selected once). But it's not so bad, let's leave it as a little improvement to be done before the official release :)

image

supabase/functions/providers.ts
```diff
@@ -34,6 +16,77 @@ export async function fetchNews(
     provider: Provider,
     newsSettings: NewsSettings,
   ): Promise<News[]> {
+    async function innerFetch(
+      provider: Provider,
+      topics: string[],
+      topic: string | undefined,
+    ) {
+      let url = provider.replaceAll(":query", topic || "");
+      if(url.startsWith("/")) {
```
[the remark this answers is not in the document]
pmdlt: Done

supabase/functions/providers.ts
```diff
@@ -34,6 +16,77 @@ export async function fetchNews(
...
+          },
+        };
+      })
+        .filter((i: News) => (Date.now() - i.publishedAt.getTime()) < 86400000);
+
+      if (topic) {
+        return news;
+      }
+
+      // Filter news by AI according to user settings.
```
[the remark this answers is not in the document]
pmdlt: Done

PR #139, opened 2024-05-08
136 feat add transcript news sharing options

pmdlt:
This PR enhances the user experience of sharing news transcripts by adding expanded share options within the app. The main changes introduced include:

1. **Expandable Share Button**: A new expandable Floating Action Button (FAB) is implemented to accommodate multiple sharing options for news transcripts.

2. **Share Options**: Three share options are provided within the expandable FAB:
   - **Text Transcript**: Users can share the full text transcript of the news article.
   - **Audio Transcript**: Users can share an audio file of the news transcript. The audio file is generated dynamically and stored locally.
   - **Link to Article**: Users can share a direct link to the news article.

**Implementation Details**:

- **ExpandableFab**: This class manages the expandable FAB functionality, including animation for opening and closing the FAB and arranging the share options around it.
- **_ExpandingActionButton**: This widget handles the animation and positioning of individual share options based on the expansion state of the FAB.
- **ActionButton**: Represents a single share option button with an icon.

Capture d'écran 2024-05-17 203904
Capture d'écran 2024-05-17 203859


pmdlt:
@Thechi2000 I'm sorry to bother you, but I don't understand why your test doesn't work.
`/home/runner/work/app/app/test/integration/sources.dart: User can consult source articles by tapping on them (failed)`
I tested several configurations, modified the test by updating the new requests returned by the Edge function, but there was nothing I could do...

Thechi2000:
The news' view is stuck in an infinite loop, this needs to be fixed. It also blocked the testing of the new feature, so I cannot give feedback on it for now.

lib/views/news_view.dart
```diff
@@ -40,13 +44,36 @@ class _NewsViewState extends State<NewsView> {
                 "The first one will be generated the first time the alarm goes off.");
       }
     } else {
-      body = ListView.builder(
-          itemCount: _newsList.length,
-          itemBuilder: (context, index) {
-            return NewsText(news: _newsList[index]);
-          });
+      var firstTranscript = _newsList.first;
+      body = Scaffold(
```
LeanderCretegny, line 48: In my opinion we definitely should not put a scaffold inside another scaffold, but due to the short amount of time left, I suggest to leave it like that for now and open an issue to correct it in the future.

ClementJosso (approved):
Good work! I see that Leander has talked about the nested Scaffold, I think it can easily be fixed with some slight tweaking of the Master View, we can do that in another PR. Also, personal opinion but I think it would look better if the main button had the Button_theme instead of the Primary_theme.

lib/views/news_view.dart
```diff
@@ -40,13 +44,36 @@ class _NewsViewState extends State<NewsView> {
...
+          body: ListView.builder(
+              itemCount: _newsList.length,
+              itemBuilder: (context, index) {
+                return NewsText(news: _newsList[index]);
+              }),
+          floatingActionButton: ExpandableFab(
+            distance: 112,
```
LeanderCretegny, line 55: Try to avoid raw number, it is better practice to declare a const variable as we did for the theme colors

Thechi2000 (commented):
Awesome ! This works flawlessly, although I've a small complain:

If I understand correctly, only the first transcript can be shared ? This seems very user-unfriendly, especially since the button stays while scrolling down.

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+  late final Animation<double> _expandAnimation;
+  bool _open = false;
+
+  @override
+  void initState() {
+    super.initState();
+    _open = widget.initialOpen ?? false;
+    _controller = AnimationController(
+      value: _open ? 1.0 : 0.0,
+      duration: const Duration(milliseconds: 250),
```
LeanderCretegny, line 33: Same thing as above, avoid magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+          _buildTapToOpenFab(),
+        ],
+      ),
+    );
+  }
+
+  Widget _buildTapToCloseFab() {
+    return SizedBox(
+      width: 56,
+      height: 56,
```
LeanderCretegny, line 78: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+      child: Center(
+        child: Material(
+          shape: const CircleBorder(),
+          clipBehavior: Clip.antiAlias,
+          elevation: 4,
```
LeanderCretegny, line 83: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+          child: InkWell(
+            onTap: _toggle,
+            child: Padding(
+              padding: const EdgeInsets.all(8),
```
LeanderCretegny, line 87: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+          ),
+        ),
+      ),
+    );
+  }
+
+  List<Widget> _buildExpandingActionButtons() {
+    final children = <Widget>[];
+    final count = widget.children.length;
+    final step = 90.0 / (count - 1);
```
LeanderCretegny, line 102: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+
+  Widget _buildTapToOpenFab() {
+    return IgnorePointer(
+      ignoring: _open,
+      child: AnimatedContainer(
+        transformAlignment: Alignment.center,
+        transform: Matrix4.diagonal3Values(
+          _open ? 0.7 : 1.0,
+          _open ? 0.7 : 1.0,
+          1.0,
```
LeanderCretegny, line 126: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+        ),
+        duration: const Duration(milliseconds: 250),
```
LeanderCretegny, line 128: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+        curve: const Interval(0.0, 0.5, curve: Curves.easeOut),
+        child: AnimatedOpacity(
+          opacity: _open ? 0.0 : 1.0,
+          curve: const Interval(0.25, 1.0, curve: Curves.easeInOut),
+          duration: const Duration(milliseconds: 250),
```
LeanderCretegny, line 133: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+    return AnimatedBuilder(
+      animation: progress,
+      builder: (context, child) {
+        final offset = Offset.fromDirection(
+          directionInDegrees * (math.pi / 180.0),
+          progress.value * maxDistance,
+        );
+        return Positioned(
+          right: 4.0 + offset.dx,
+          bottom: 4.0 + offset.dy,
```
LeanderCretegny, line 169: Magic number

lib/views/master_view.dart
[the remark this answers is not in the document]
pmdlt: I'll take care of it! Thanks :)

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+  final Widget icon;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    return Material(
+      shape: const CircleBorder(),
+      clipBehavior: Clip.antiAlias,
+      color: theme.colorScheme.secondary,
+      elevation: 4,
```
LeanderCretegny, line 202: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+    required this.isBig,
+  });
+
+  final bool isBig;
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      margin: const EdgeInsets.symmetric(vertical: 8, horizontal: 24),
+      height: isBig ? 128 : 36,
```
LeanderCretegny, line 225: Magic number

lib/widgets/share_button.dart
```diff
@@ -0,0 +1,232 @@
...
+      decoration: BoxDecoration(
+        borderRadius: const BorderRadius.all(Radius.circular(8)),
+        color: Colors.grey.shade300,
```
LeanderCretegny, line 228: Please use the theme defined lib/utils. If you need another color you can add it there.

LeanderCretegny (changes requested):
Apart from the magic number and the scaffold thing the code is very nice. I request change to remove the magic numbers, we'll deal with the scaffold another time.

pmdlt:
> Awesome ! This works flawlessly, although I've a small complain:
> 
> If I understand correctly, only the first transcript can be shared ? This seems very user-unfriendly, especially since the button stays while scrolling down.

Yes, I agree, that's why I was pushing for a separate view between news alone and the list. But I think it's ok to leave it like that for now as a PoC and we'll discuss it at the next meeting :)

lib/views/news_view.dart
pmdlt: I've already thought about this, see https://stackoverflow.com/questions/60561552/multiple-scaffolds-for-each-page-inside-a-flutter-app

ClementJosso (approved):

lib/views/news_view.dart
pmdlt: I agree in principle, but for this particular use case I don't think it's appropriate to turn them all into variables. And I accept this choice. It's a module that's grafted onto the app, not a main feature.

PR #142, opened 2024-05-09
141 feat add alarm in onboarding

LeanderCretegny:
Add screen to onboarding to set up a simple first alarm when user registers for the first time on the app.

To test: Go through the onboarding and try to add the alarm, test if each button behave as expected.


lib/views/alarm_wizard.dart
```diff
@@ -0,0 +1,86 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _selectedTime = DateTime.now()
+        .add(const Duration(minutes: 5))
+        .copyWith(second: 0, millisecond: 0);
+    loopAudio = true;
+    vibrate = true;
+    volume = 0.3;
+    assetAudio = "assets/audio/boom.mp3";
```
quartztz, line 32: wouldn't be against having a way to choose your own sound right around here ! nothing for now, just food for thought as to where we would implement that :D

lib/views/alarm_wizard.dart
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    AlarmsViewModel avm = Provider.of<AlarmsViewModel>(context);
+    AuthModel auth = Provider.of<AuthModel>(context);
+
+    Widget body = Column(
+      mainAxisAlignment: MainAxisAlignment.spaceEvenly,
```
quartztz, line 41: I understand why this was done : it's fast to implement and lays everything out nicely. I just don't know if it fully fits the design language of our application. Can and will this be discussed IRL ? Yes ! Can't wait to talk about it with you :)

lib/views/alarm_wizard.dart
```diff
@@ -0,0 +1,86 @@
...
+      children: [
+        Text("Choose your first alarm!",
+            style: Theme.of(context).textTheme.titleLarge),
+        PickTimeButton(
+            initialTime: _selectedTime,
+            onTimeSelected: (time) {
+              _selectedTime = time;
+            }),
+        WizardNavigationBottomBar(
```
quartztz, line 50: ...especially because it breaks the display language for the "bottom bar" becoming a "bottom-ish bar we won't talk about". (follows from the previous comment, sorry)

lib/widgets/alarms_widget.dart
```diff
@@ -0,0 +1,71 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return RawMaterialButton(
+      onPressed: pickTime,
+      fillColor: Colors.grey[200],
+      shape: const RoundedRectangleBorder(
+          borderRadius: BorderRadius.all(Radius.circular(10))),
+      child: Container(
+        margin: const EdgeInsets.all(20),
```
quartztz, line 60: Please use the values provided by the Theme I made ! You've already done it for the text, there should be all the colors and spacings you need to make the app feel more cohesive. If there aren't, feel free to add ! But do check whether or not they're too similar to already existing values.

quartztz: I'll request changes now for this, sorry, but it's fast enough and it will be easier to do now rather than later. 

_Also, it's not like this will be merged into main before the meeting,,,_

quartztz (changes requested):
See my comment about themes for details about the requested changes ! Sorry >.<

lib/views/alarm_wizard.dart
LeanderCretegny: I think this is out of the scope of this PR, but is not a bad idea. I would suggest to talk about it with everyone, and in case it is validated open a new PR.

lib/views/alarm_wizard.dart
LeanderCretegny: Yes no problem, but please next time try to do these comment sooner as I did asked several time for feedback on the telegram group, but had no answer :'(.

lib/widgets/alarms_widget.dart
LeanderCretegny: Yes sorry, this code is in fact a refactor of a widget made by someone else for the alarm settings, and I admit that I did not really take a look at the details, I will do the requested changes asap

lib/views/alarm_wizard.dart
pmdlt: Maybe the user doesn't want to bother with it in unboarding, but it definitely needs to be included in the settings.

quartztz (dismissed):

lib/widgets/alarms_widget.dart
```diff
@@ -0,0 +1,69 @@
...
+
+    if (res != null) {
+      setState(() {
+        final now = DateTime.now();
+        _selectedTime = now.copyWith(
+          hour: res.hour,
+          minute: res.minute,
+          second: 0,
+          millisecond: 0,
+          microsecond: 0,
```
Pacsort17, line 42: Why 0 seconds / milliseconds / microseconds ?

Pacsort17 (changes requested):
During the onboarding, the navigation loops infinitely (when pressing next) over the "Choose your interests" and "Choose your provider" views

lib/views/news_alert_setup_view.dart
```diff
@@ -163,6 +127,7 @@ class _NewsAlertSetupViewState extends State<NewsAlertSetupView> {
                             else
                               {deleteAlarm(context)}
                           },
+                          key: const Key("Switch-on-off"),
```
Thechi2000, line 130: Until now, we've used "kebab-case" for keys, it would be good practice to continue with the same.

lib/views/news_alert_setup_view.dart
```diff
@@ -227,6 +192,7 @@ class _NewsAlertSetupViewState extends State<NewsAlertSetupView> {
                         loopAudio = value;
                         updateAlarm(context);
                       }),
+                      key: const Key("Switch-loop"),
```
Thechi2000, line 195: Same as above: Until now, we've used "kebab-case" for keys, it would be good practice to continue with the same.

lib/views/news_alert_setup_view.dart
```diff
@@ -243,6 +209,7 @@ class _NewsAlertSetupViewState extends State<NewsAlertSetupView> {
                         vibrate = value;
                         updateAlarm(context);
                       }),
+                      key: const Key("Switch-vibrate"),
```
Thechi2000, line 212: Same as above: Until now, we've used "kebab-case" for keys, it would be good practice to continue with the same.

lib/views/providers_wizard_view.dart
Thechi2000, line 1: Thanks for the refactor and fixes !

lib/widgets/alarms_widget.dart
```diff
@@ -0,0 +1,69 @@
...
+            name: "Debug", level: Level.WARNING.value);
+      });
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return RawMaterialButton(
+      onPressed: pickTime,
+      fillColor: THEME_BUTTON, //Colors.grey[200],
```
Thechi2000, line 57:
```suggestion
      fillColor: THEME_BUTTON,
```

Not sure about the comment here, it can be misleading especially if we change the value of the constant later on.

lib/widgets/alarms_widget.dart
```diff
@@ -0,0 +1,69 @@
+import 'dart:developer';
+import 'dart:ffi';
```
Thechi2000, line 2:
```suggestion
```

This import is unused.

lib/main.dart
```diff
@@ -4,6 +4,7 @@ import 'package:actualia/models/auth_model.dart';
 import 'package:actualia/viewmodels/providers.dart';
 import 'package:actualia/utils/themes.dart';
 import 'package:actualia/viewmodels/alarms.dart';
+import 'package:actualia/views/alarm_wizard.dart';
```
Thechi2000, line 7:
```suggestion
```

The import is unused.

test/widgets/profile_view.dart
```diff
@@ -203,4 +209,32 @@ void main() {
...
+    await tester.tap(find.byType(PickTimeButton));
+    await tester.pumpAndSettle();
+    await tester.tap(find.byIcon(Icons.keyboard_outlined));
+    await tester.pumpAndSettle();
+    await tester.tap(find.text("OK"));
+
+    await tester.tap(find.byKey(const Key("Switch-on-off")));
+    await tester.drag(find.byType(Slider), Offset.fromDirection(30));
+    await tester.tap(find.byKey(const Key("Switch-loop")));
+    await tester.tap(find.byKey(const Key("Switch-vibrate")));
```
Thechi2000, line 233: Same as above: Until now, we've used "kebab-case" for keys, it would be good practice to continue with the same.

Thechi2000 (commented):
Awesome job, thank you so much ! The code is really clean (just a few nit pick as usual), and the UI is good. Tested on Android.

I just have one issue with the UI: the "Alarm is set for tomorrow" and "Alarm not set" seems to be randomly updated when using the different controls, which is really not UX friendly. I'm also not sure about the point of it, since it will be set when tapping on "Done", right ?

LeanderCretegny:
> I just have one issue with the UI: the "Alarm is set for tomorrow" and "Alarm not set" seems to be randomly updated when using the different controls, which is really not UX friendly. I'm also not sure about the point of it, since it will be set when tapping on "Done", right ?

This is indeed an issue, but that has nothing to do with this PR. This PR is only about the onboarding, not the settings.

Thechi2000 (approved):
Nice, let's merge that :)

Pacsort17 (approved):
Tested on Android, the alarm can be setup during the onboarding (and it rings), thanks for the job !

PR #145, opened 2024-05-10
144 Take pictures from android

ClementJosso:
This PR adds the possibility to take pictures of text you see IRL with the app. This text is then read with OCR and sent to an LLM to provide context (this was implemented in [PR 149]( https://github.com/ActualIA/app/pull/149)). This corresponds to our 3rd epic.

Specifically, this PR allows the user to take pictures, and uses the new navigation bar from [PR 148](https://github.com/ActualIA/app/pull/148) as well as the VM from PR 149 to display the context obtained from the picture.

The UI view for the context will be added in another PR, so as of right now the view is a simple text widget.
In addition, this PR removes some of the warning we had in files throughout the app such as missing const, unused imports etc.

**Testing** the new functionality is fairly straightforward : Run the app on your smartphone, click on the camera icon on the bottom navigation bar and take a picture of some text you see (ideally news). The app should display a text with some additional context of what you just photographed.

**NB** : As of right now, if you leave the view with the context, you can't go back, as clicking on the camera icon again will make you take another picture. This is intended for this PR.

**IMPORTANT** : Do not merge this PR until PR 148 is merged


pmdlt:
I think we should merge #149 first, what do you think?

LeanderCretegny (commented):
First of all, I find the code pretty nice, nothing to change in my opinion. I particularly like the way you handle the issue of having a fourth view (the context) and keep the camera button at the center of the navigation view.
That said, I see that their is no tests for the news recognition view model, is it intended?

ClementJosso:
> I see that their is no tests for the news recognition view model, is it intended?
Yes, since I only added one function to this existing VM, I believe it would make more sense to have a separate PR dedicated to testing the VM as a whole.

LeanderCretegny:
In this case yes I agree with you, we should do a separate PR for that.

LeanderCretegny (approved):

pmdlt (approved):
Looks good from here!

PR #148, opened 2024-05-10
143 feat navigation menu

ClementJosso:


lib/views/master_view.dart
```diff
@@ -0,0 +1,78 @@
...
+  State<MasterView> createState() => _MasterView();
+}
+
+class _MasterView extends State<MasterView> {
+  Views _currentViews = Views.NEWS;
+  late List<Destination> _destinations;
+
+  void setCurrentViewState(Views view) {
+    if (view == Views.CAMERA) {
+      log("Camera button pressed on navigation bar",
```
pmdlt, line 24: Why add these logs? Is there any point in putting them in production?
They'll have to be removed for the final version of the app anyway.

ClementJosso: I agree, these logs will have to be removed, but this is currently a placeholder, because a future PR will replace this part with an actual action upon clicking on the camera button. In my opinion we can leave it for now as long as it gets removed by the PR implementing the camera.

pmdlt (approved):
Thanks for your replies!

Thechi2000:
Code looks great ! However, I'm not convinced by the UI, I think it would be good to have a clear delimitation for the navbar (e.g. using a slightly different background color), and the icons could be bigger and centered (with spacers), because it feels really weird IMO.

image

Thechi2000 (approved):
LGTM ! Good work :)

pmdlt (approved):
LGTM :)

PR #151, opened 2024-05-13
150 providers view input sanitization

### LINKED ISSUE #150: Provider's view input sanitization
The different fields in the providers' wizard view should be checked to be valid before uploading to the database, and the subsequent error message(s) should be displayed.

Thechi2000:
Adds provider's parameters validation through the use of a factory, put in the `ProviderType`. It returns either the `NewsProvider` or the a list of error, for each fields.

Also, I removed all the debug prints polluting the output.

Closes #150


quartztz:
Thank you for cleaning prints ! I'll test on device in a few minutes, but at the very least I can comment on how useful that is :D 
Also, clean code ! Super clear, I wish all reviews were this straightforward to read through <3

quartztz:
I have some comments about UI and UX ! I'll show them in the code whenever I can. But overall, functionality is great and seems to work really well ! Thanks for the PR

lib/viewmodels/alarms.dart
```diff
@@ -60,10 +58,9 @@ class AlarmsViewModel extends ChangeNotifier {
           'timetz': timetz,
           'transcript_settings_id': settingsId
         }, onConflict: "created_by");
-        print("Alarm pushed on supabase");
       } catch (e) {
         //log("Error pushing alarm: $e", level: Level.WARNING.value);
```
quartztz, line 62: nitpick: if line 63 is merged, this one could be removed :D

lib/views/providers_wizard_view.dart
```diff
@@ -68,8 +68,10 @@ class _ProvidersWizardView extends State<ProvidersWizardView> {
           onCancel: () => Navigator.pop(context),
           rOnPressed: () async {
             pvm.updateProvidersFromEdited();
-            if (!await pvm.pushNewsProviders() && Platform.isAndroid) {
-              Fluttertoast.showToast(msg: "Error while updating providers");
+            if (!await pvm.pushNewsProviders()) {
+              if (Platform.isAndroid) {
+                Fluttertoast.showToast(msg: "Error while updating providers");
+              }
```
quartztz, line 74: Is there a reason to split this `&&` into two successive `if`s? Unless I misunderstand this code, it seems to me like the functionality is the same.

lib/widgets/wizard_widgets.dart
```diff
@@ -190,10 +190,17 @@ class _ProviderWidgetState extends State<ProviderWidget> {
...
-    var fields = type.parameters.map((e) {
-      var index = type.parameters.indexOf(e);
-      return TextField(
+    var fields = type.parameters.indexed.expand((el) {
+      var index = el.$1;
+      var e = el.$2;
+
+      var error = errors?[index];
+      var erroredTheme = const TextStyle(
+          color: Color.fromARGB(255, 163, 0, 0), fontWeight: FontWeight.bold);
```
quartztz, line 201: If only there was a way to save themes as constants...
...
...

`themes.dart`

lib/widgets/wizard_widgets.dart
```diff
@@ -203,6 +210,20 @@ class _ProviderWidgetState extends State<ProviderWidget> {
...
           });
+
+      var errorMessage = error != null
+          ? Container(
+              padding: const EdgeInsets.fromLTRB(UNIT_PADDING, 0, 0, 0),
+              child: Text(error,
+                  style: Theme.of(context)
+                      .textTheme
+                      .labelLarge
+                      ?.merge(erroredTheme)))
```
quartztz, line 221: At least you treated the theme merging well ! I still think it would be more appropriate to have error themes in the main theme file, since they might (and will, hopefully) be reused throughout the app. At least for the color !

quartztz (changes requested):
Comments have been left, mostly on the stylistic choices both on the app and in the code. An additional comment, which might be beyond the scope, but which interrogates the behavior of the provider card : is it possible for the validation of the input to happen not on full component confirm, but on text input validation ? I think it would make more sense to alert the user of a mistake __before__ they confirm, but it might require some rework, as I don't see how it could simply plug into the code as of right now. Still,

lib/views/providers_wizard_view.dart
Thechi2000: Yes, it is to avoid the else cases. We don't want to pop the navigator if the push failed.

lib/widgets/wizard_widgets.dart
Thechi2000: Is it ok for you this way ?

quartztz (dismissed):

quartztz (approved):
Thanks for addressing the comments ! Looks good for now.

LeanderCretegny (approved):
Code is great, and I could not found a bug while testing on the emulator

PR #153, opened 2024-05-13
152 feat add backend for raw rss feed & add providers

pmdlt:
> EDIT: Smaller PR than planned following our discussion last Tuesday, we will refactor the choice of providers to use both views (those of @quartztz  and @Thechi2000).
> **This PR no longer includes the addition of new providers**

This PR implements the backend functionality for raw RSS feed support ~~and adds new providers to the list of available sources~~. 

Changes:
- Added backend functionality to generate raw RSS feed.
- ~~Updated `lib/models/providers.dart` to include new provider types.~~
- Created `generate-raw-feed` function in `supabase/functions` directory to handle RSS feed generation.
- Updated `supabase/functions/generate-raw-feed/index.ts` to fetch user interests and fetch news from selected providers.

This enhancement expands the capabilities of the application by allowing users to fetch news ~~from additional sources~~ via raw RSS feeds.


Thechi2000:
I think it would be more interesting to have a new category for predefined feeds, to avoid the user having to sift through all of them when adding new providers. What do you think ?

pmdlt:
> I think it would be more interesting to have a new category for predefined feeds, to avoid the user having to sift through all of them when adding new providers. What do you think ?

Hello @Thechi2000!
Yes, so I propose to do all that in another PR, since it might be a big job to refactor the whole providers part to do what we discussed last Tuesday.
So I'll withdraw my changes to the providers, and re-implement all that in a second time, ok for you?

Pacsort17 (approved):
Thanks for the job !

quartztz (approved):
Changes look great to me ! They'll be useful in implementing my view :) Thank you for the work!

PR #156, opened 2024-05-14
fix: fix body sent for generate-transcript

Thechi2000:
Fixes the call to `generate-transcript`, since it requires a JSON body.

### DIFF
--- a/lib/viewmodels/news.dart
+++ b/lib/viewmodels/news.dart
@@ -169,7 +169,7 @@ class NewsViewModel extends ChangeNotifier {
   /// Invokes a cloud function to generate news transcripts.
   Future<void> invokeTranscriptFunction() async {
     try {
-      await supabase.functions.invoke('generate-transcript');
+      await supabase.functions.invoke('generate-transcript', body: {});
       log("Cloud function 'transcript' invoked successfully.",
           level: Level.INFO.value);
     } catch (e) {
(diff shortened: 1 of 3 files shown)


pmdlt (dismissed):

lib/viewmodels/news.dart
```diff
@@ -85,6 +85,8 @@ class NewsViewModel extends ChangeNotifier {
   }
 
   Future<void> getNewsList() async {
+    print("suzdgfshidufh");
```
quartztz, line 88: incredible work on the PR, but You Are Not Above Stylistic Guidelines. 

please remove stray prints. 

or else.

test/integration/sources.dart
```diff
@@ -183,7 +183,7 @@ class MockHttp extends BaseMockedHttpClient {
                 },
               ]
             },
-            "date": "2024-04-30T16:48:58.691625+00:00",
+            "date": DateTime.now().toIso8601String(),
```
quartztz, line 186: Can't believe this was the source of the bug...

Oh well ! Glad it's fixed now :D

quartztz (changes requested):
Incredible PR ! However, not above style guides. Please revise quickly! It's blocking for many other PRs.

lib/viewmodels/news.dart
Thechi2000: oopsie, thanks

quartztz (approved):

pmdlt (approved):

PR #157, opened 2024-05-15
107 fix title of daily transcripts is always the same

pmdlt:
Small PR that adds AI generation of the news title (via the LL-submodel, when creating the JSON).

Note: the title is now both in the JSON transcript and in the `title` row. Ideally, we'd like to avoid duplication in the DB, but for the sake of backward compatibility, I suggest leaving it like this for now.


ClementJosso (approved):
This is great news ! Thanks for the work

quartztz:
Nice QOL/UX change ! Code looks great as well :D

quartztz (approved):

PR #164, opened 2024-05-17
160 Implement raw RSS view 

quartztz:
According to our new user story, we want to be able to inspect every news fed to the LLM, as if it were an RSS feed. This view implements just that ! 

For now, lacks a ViewModel to plug into the backend. Coming today !


lib/viewmodels/rss_feed.dart
```diff
@@ -0,0 +1,50 @@
...
+  @protected
+  void setArticles(List<Article> articles) {
+    _articles = articles;
+  }
+
+  List<Article> get articles => _articles ?? [];
+
+  RSSFeedViewModel(this.supabase);
+
+  geRawNewsList() async {
```
LeanderCretegny, line 19: This looks like a small typo

lib/views/rss_feed_view.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:logging/logging.dart';
+import 'package:provider/provider.dart';
+
+// TODO: Implement ViewModel plugging into existing EF.
+
+class FeedView extends StatefulWidget {
+  const FeedView({super.key});
+
+  @override
+  State<FeedView> createState() => _FeedViewState();
```
LeanderCretegny, line 18: small nitpick, until now and as far as I know, the naming conventions we followed did not bother adding the extra State at the end of the class name, _FeedView is enough.

lib/views/rss_feed_view.dart
LeanderCretegny, line 1: I think this view could be Stateless, it does not seem to hold any state.

lib/views/source_view.dart
```diff
@@ -6,40 +7,34 @@ import 'package:url_launcher/url_launcher.dart';
...
+  final Article article;
 
-  const SourceView(
-      {this.article = "",
-      this.title = "",
-      this.date = "",
-      this.newsPaper = "",
-      this.url = "",
-      super.key});
+  const SourceView({this.article = const Article(), super.key});
```
LeanderCretegny, line 12: Wouldn't it make more sense to set the article parameter as required instead of providing a default one?

lib/views/source_view.dart
LeanderCretegny, line 1: I don't think it is in the scope of this PR but it is worth mentioning, it seems to me that the SourceView could also be a stateless widget

LeanderCretegny (changes requested):
Overall pretty great, a lot of small refactor that are much appreciated, my only concerns are about some view that imo could be stateless widget instead of statefull ones.

lib/views/rss_feed_view.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:flutter/material.dart';
+import 'package:logging/logging.dart';
+import 'package:provider/provider.dart';
+
+class FeedView extends StatelessWidget {
+  const FeedView({super.key});
+
+  Widget build(BuildContext context) {
+    Widget loading =
+        const LoadingView(text: "Please wait while we fetch the news for you.");
```
LeanderCretegny, line 17: Shouldn't it be localized?

lib/widgets/sources_view_widgets.dart
```diff
@@ -41,3 +47,68 @@ class SourceTitle extends StatelessWidget {
...
+  final String content;
+  final String sourceUrl;
+
+  const ArticleWidget(
+      {this.title = "",
+      this.origin = "",
+      this.date = "",
+      this.content = "",
+      this.sourceUrl = "",
+      super.key});
```
LeanderCretegny, line 64: Wouldn't it be better to use the Article class and have only one attribute instead of 5?

lib/widgets/sources_view_widgets.dart
```diff
@@ -41,3 +47,68 @@ class SourceTitle extends StatelessWidget {
...
+
+  @override
+  Widget build(BuildContext context) {
+    var loc = AppLocalizations.of(context)!;
+
+    return ListView(
+      shrinkWrap: true,
+      physics: const NeverScrollableScrollPhysics(),
```
LeanderCretegny, line 72: Please use ListView only when you need a scrollable list, consider using Column when it is not the case.

lib/widgets/sources_view_widgets.dart
```diff
@@ -41,3 +47,68 @@ class SourceTitle extends StatelessWidget {
...
+      children: <Widget>[
+        ListView(
+            shrinkWrap: true,
+            physics: const NeverScrollableScrollPhysics(),
```
LeanderCretegny, line 76: Same comment as above

lib/widgets/sources_view_widgets.dart
LeanderCretegny, line 1: Upon testing on the emulator I found one weird thing. The widget `ScrollableText` does not seem to be scrollable, I think this is due to the text being to short to be scrolled, but then why bother with creating this widget, maybe it is plan in the future to have bigger text?

LeanderCretegny (changes requested):
Overall good code with much appreciated refactor, except in the ArticleWidget widget for which I am asking some changes about the unnecessary use of ListView instead of Column.

lib/main.dart
```diff
@@ -72,19 +74,22 @@ class _AppState extends State<App> {
...
     } else if (authModel.isSignedIn) {
-      newsSettings = Provider.of(context);
+      newsSettingsModel = Provider.of(context);
       if (authModel.isOnboardingRequired) {
-        if (newsSettings.settings == null || pvm.newsProviders == null) {
-          home = LoadingView(text: AppLocalizations.of(context)!.loading);
+        if (newsSettingsModel.settings == null ||
+            providersModel.newsProviders == null) {
+          // TODO: this is not normal :( ;
+          home = const LoadingView(text: 'Fetching your settings...');
```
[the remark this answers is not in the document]
quartztz: Whoops ! not sure how that happened, i never had it on my machine. At any rate, that can be changed quickly !

lib/views/rss_feed_view.dart
quartztz: Right ! Missed that one.

LeanderCretegny (approved):

lib/widgets/sources_view_widgets.dart
```diff
@@ -41,3 +46,55 @@ class SourceTitle extends StatelessWidget {
...
+
+  const ArticleWidget({required this.article, super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    var loc = AppLocalizations.of(context)!;
+
+    return ListView(
+      shrinkWrap: true,
+      children: <Widget>[
```
LaClimSx, line 61: I think this should have NeverScrollable physics, since this listView will be nested in another one

LaClimSx (changes requested):

quartztz:
Thanks for handling it Clément !

LaClimSx (approved):
The scroll works now, and apart from that, the RSS view is fully functional, thanks for that!

LeanderCretegny (approved):

PR #178, opened 2024-05-24
177 implement narrator settings

LaClimSx:
This PR implements a new setting, the customization of the narrator (i.e. the voice that reads the transcripts in the audio file). This is seen by clicking on the narrator view in the profile page. As far as my testing goes, the view and state work well, and checking an option does change the value in the DB. However, when generating an audio with the correct voice in the DB, it still uses the default voice. I believe this problem does not come from the code of this PR but from the way we use the DB parameter in the generate-audio edge function (Will have to check with people more knowledgeable than me on that part). 

**For reviewers**: The main parts to assert from you is that the view is good looking, that the options work and save state, and that you can play the audio demo of a voice when clicking on a play button.

PS: you might have noticed there are no automated tests for this yet, this in intended, I will add them soon.


Thechi2000 (commented):
Nothing to say about the code, it's very clean, and follows the good practice guidelines.

About the UI, I think we should have the bottom bar with the `Cancel` and `Done` buttons in the narrator setttings, but that's all.

Great work overall !

LaClimSx:
> About the UI, I think we should have the bottom bar with the `Cancel` and `Done` buttons in the narrator setttings, but that's all.

I actually didn't add these buttons on purpose, because I believe they are not really necessary. In practice a done button would do nothing more than going back to the previous view, since the settings are saved each time you click on an option. It is true however that the app does have done buttons pretty much everywhere, so for the sake of consistency we can add one, but I'm not so sure about a cancel button, since you only need one touch to revert your settings to the voice you have before.

Thechi2000:
I think having a `Done` button would be good for consistency, as you said. It does not need to be doing anything more than going back to the previous screen, imo. As you pointed it out, a `Cancel` button would not add much since the changes can very easily reverted by hand.

LeanderCretegny (changes requested):
By running the code on an emulator I have the following two remarks : 
- First a UI detail: I think it would look better if the list would be thinner (more precisely less large than the title).
- Secondly, when changing my selection, there is an exception caught by my IDE. It does not crash the app but I think it is not a good think to have unmanaged error like that in the code. 
image

Small remark concerning the done and cancel button: imo indeed a back button is not required, but the done is a good think in terms of UX. Also please (when it will be merged, right now it is still waiting for validation) use the new bottom bar.

Thechi2000 (approved):
Looks good to me :)

LeanderCretegny (changes requested):
You did not use the bottom bar :,(. Also I'm sorry I did not seen it before but you should also remove the user button in the top bar. 
image

Thechi2000 (approved):

LeanderCretegny (approved):
All requested changes have been done, and everything works fine on my emulator!

PR #187, opened 2024-05-26
fix: fix bug in onboarding

### LINKED ISSUE #186: fix: major bug in onboarding
The onboarding is not closing correctly, instead of landing on the main pages, the app is stuck in an invalid state that should not be reachable.

LeanderCretegny:
Fixes the bug in the onboarding that was breaking the app flow. Upon completion the onboarding was landing on an invalid state instead of the main view.

Testing: run the ci and go through the onboarding in a phone or emulator to check that the app flow is correct.

closes #186


Thechi2000 (approved):
Nice catch ! Tested on Android with a fresh new account, works flawlessly.

pmdlt:
Thanks for the fix

pmdlt (approved):
Thanks for the fix

PR #189, opened 2024-05-26
feat: replace guest login with anonymous sign in option

quartztz:
This pull request will modify the codebase's approach to anonymous/temporary users by using the existing [Anonymous Sign-In API](https://supabase.com/blog/anonymous-sign-ins) that's made available by Supabase Auth. This is a cleaner solution, as it allows users to use the app as intended without a Google Sign-In, with their own interests and so on, without impacting others in the same situation. 

Still in drafts !

A few things must be noted. 
> ℹ️  I had to modify the settings in the Supabase dashboard to explicitly allow Anonymous Sign-Ins. Be mindful of that !

> ⚠️ This inevitably opens us to security concerns that we didn't have earlier, such as potential botting practices or other exotic attacks through anonymous users. We might want to see how this interacts with our RLS, or if we want to add some sort of Captcha anti-bot behavior.


Thechi2000 (approved):
Awesome ! This will be useful both for showcasing, testing and production. Nothing to say about the code, it is clean and concise.

lib/models/auth_model.dart
```diff
@@ -90,8 +90,14 @@ class AuthModel extends ChangeNotifier {
     return res.user != null;
   }
 
+  // use the anonymous sign in API from Supabase !
+  // More info can be found at the documentation webpage.
   Future<bool> signInWithFakeAccount() async {
```
pmdlt, line 95:
```suggestion
  Future<bool> signInAnonymously() async {
```

pmdlt (changes requested):
Thanks for the work!

I think there are still 1-2 small things to do, including removing all traces of the old login form, and why not (I'll let you judge if it's relevant or not) indicate in the ui that the “sign in as a guest” is local, like : `Sign in as a guest (local account)` or something

lib/models/auth_model.dart
pmdlt: Thanks for the changes!

pmdlt (approved):
Thanks for the changes!

pmdlt (commented):
I merged too quickly, the requested changes were not made.

PR #194, opened 2024-05-29
192 database integration ux

### LINKED ISSUE #192: Improve UX for backend integration
The integration with the backend is still lacking critical UX features, such as:

Error handling. When an error occurs, the user should be correctly notify, either by a toast or by a persistent message on screen. This includes:

- [x] failing to fetch news
- [x] failing to update settings

Loading. Any view using the database should include a loading screen for when the operation takes too much time, to avoid confusing the user:
- [x] updating interests settings
- [x] sending image for the OCR

Thechi2000:
This PR aims to improve error and loading screens for better user experience. To do so, I refactored the `NewsViewModel` to clean it up. 

I also made major changes to the OCR handling: it is now handled from the `ContextView`, and the `MasterView` always show it instead of managing the OCR workflow. When the user enters the `ContextView` for the first time, or when they it the floating button, the workflow is started, first prompting them to take a picture, etc.

Closes #192.


pmdlt (commented):
A few ideas for changes related to the French translation

Pacsort17 (dismissed):
Tested on Android, everything works.

Nice refactor and integration, I have nothing to say :)

lib/views/context_view.dart
```diff
@@ -1,34 +1,63 @@
...
+      child: Scaffold(
+          floatingActionButton: IconButton.filledTonal(
+            iconSize: 40,
+            onPressed: () => nrvm.takePictureAndProcess(),
+            icon: Container(
+                padding: const EdgeInsets.all(UNIT_PADDING / 4),
+                child: const Icon(Icons.camera_alt)),
+            color: THEME_BUTTON,
+          ),
+          body: body),
```
LaClimSx, line 60: I don't know if this is a problem but here you are creating nested Scaffolds, there is a Scaffold containing a Container containing a Scaffold. It doesn't seem the be causing an issue in the app but since we talked about his on previous PRs I just want to make sure you know what you're doing.

lib/views/context_view.dart
```diff
@@ -1,34 +1,63 @@
...
+    return Container(
+      padding: const EdgeInsets.all(UNIT_PADDING),
+      child: Scaffold(
+          floatingActionButton: IconButton.filledTonal(
+            iconSize: 40,
+            onPressed: () => nrvm.takePictureAndProcess(),
+            icon: Container(
+                padding: const EdgeInsets.all(UNIT_PADDING / 4),
+                child: const Icon(Icons.camera_alt)),
+            color: THEME_BUTTON,
```
LaClimSx, line 58: I don't know if this is a problem on my emulator only, but I think the floating button looks a bit weird, the outline is always visible instead of only when hovering over it, is it the case for you too ?

lib/viewmodels/news_recognition.dart
```diff
@@ -64,6 +108,19 @@ class NewsRecognitionViewModel extends ChangeNotifier {
...
+      setError(_Error.noImage);
+      return;
+    }
+
+    markProcessingAs(true);
+    final text = await recognizeText(picture.path);
+    await invokeProcessImage(text);
+    markProcessingAs(false);
+  }
+
```
LaClimSx, line 123: I'm not sure something can be done to fix this, but upon clicking the camera button, there's a small delay before the camera actually opens where we can see an error screen.

LaClimSx (commented):
I have written a few comments but apart from that this is great work! I really like your use of "Either" to simplify the handling of errors.

lib/viewmodels/news_recognition.dart
Thechi2000: I've got the same issue, but I wasn't able to find the its origin.

lib/views/context_view.dart
Thechi2000: It is intended. The share button already implemented does this the same way.

lib/views/context_view.dart
Thechi2000: I've used the same method as `NewsView` uses to have the floating button. It seems like a good option since the `Scaffold` is the only element growing with the size of the parent (unlike the `TopAppBar` and the `NavigationBar`).

Pacsort17 (approved):

pmdlt (approved):
Still ok

PR #196, opened 2024-05-29
195 fix sticky settings in user view

### LINKED ISSUE #195: fix: sticky settings in user view
The setting views keep locally the change made even upon closing app or cancelling, which should not be the case.

LeanderCretegny:
Fix the cancel button on the settings view for interests and news providers.

**Testing**: Run the code on an emulator or real android phone, and go to the settings and try modifying them in any way you can think of and check that the cancel buttons indeed cancel the changes you made

Closes #195


Thechi2000 (approved):
Nice fix, nothing to say about the code !

Pacsort17 (commented):
Tested on Android, it works well ! Nothing to say about the code, however when clicking on the cancel button it leads the user back to the previous page. I think it should only unselect the buttons, but let's discuss that during the meeting, and then I'll approuve the PR.

PR #198, opened 2024-05-30
Augment interests & cities lists

Thechi2000:
Adds ~700 cities and ~250 interests to the selectors in the settings. 

For the cities: this [dataset](https://public.opendatasoft.com/explore/dataset/geonames-all-cities-with-a-population-1000/table/?disjunctive.cou_name_en&sort=name), then filtered to keep only the three biggest cities per country (relative to their population count).

For the interests: this [list](https://interestexplorer.io/facebook-interests-list/) was used.


LeanderCretegny (approved):
Not a lot thing to say besides thanks for taking care of this annoying task!

pmdlt (approved):

PR #199, opened 2024-05-30
feat: add filter bar to interests setting views

LeanderCretegny:
Add a filter to interests views in settings and onboarding to make searching for the wanted country, city or center of interest easier.

Testing: Run the code and try the filter by yourself and look for potential unexpected behavior. You can either go in the onboarding or in the settings to test the filter.


lib/widgets/wizard_widgets.dart
```diff
@@ -1,8 +1,11 @@
+import 'dart:developer';
```
Thechi2000, line 1:
```suggestion
```

This import is unused.

lib/widgets/wizard_widgets.dart
```diff
@@ -1,8 +1,11 @@
...
+
 import 'package:actualia/models/providers.dart';
 import 'package:actualia/viewmodels/providers.dart';
 import 'package:actualia/widgets/top_app_bar.dart';
 import 'package:actualia/utils/themes.dart';
 import 'package:fading_edge_scrollview/fading_edge_scrollview.dart';
+import 'package:flutter/cupertino.dart';
```
Thechi2000, line 8:
```suggestion
```

This import is unused.

Thechi2000 (commented):
I'm not convinced of only matching prefixes. I think it would be more convenient to the user if the search can be matched anywhere in the string, especially with the incoming new predefined lists (https://github.com/ActualIA/app/pull/198).

LeanderCretegny:
Regarding the prefix issue, you are not wrong, but there is no dart built-in function to do so (or maybe I have missed it, but I did not find one), which mean that we would need to code it ourselves and given the little amount of time we have left I would rather focus on other tasks.

Thechi2000:
> Regarding the prefix match issue, you are not wrong, but there is no dart built-in function to do so (or maybe I have missed it, but I did not find one), which mean that we would need to code it ourselves and given the little amount of time we have left I would rather focus on other tasks. 

You can use the builtin `contains()` function, if I'm not mistaken.

> For the web part, we do not really care about it honestly.

I know, I just wanted to point it out in case it was a trivial fix.

LeanderCretegny:
> You can use the builtin `contains()` function, if I'm not mistaken.
I did try, but it was not working and I incorrectly assumed the error was from the `contains()` function while it was in fact not. I have now made the change and use `contains()` instead of `matchAsPrefix()`

Thechi2000:
I just tried, it seems to work just fine. However, you need to swap the two operands otherwise you would be trying to look for the item in the filter.

```dart
where((item) => item.$2.toLowerCase().contains(_filter.toLowerCase()))
```

Pacsort17 (approved):
Tested on Android, it works ! 
The design of the box could be a smoother, but nothing important.

lib/utils/themes.dart
LaClimSx, line 1: Not something you changed on this PR, but I notice on line 230 that body can never be null so this check doesn't do anything. Maybe you could remove it ?

LaClimSx (commented):
Nothing to say on the code changed, good work! ,I've added a comment regarding existing code because it's fast to change even though you didn't technically change that part.

lib/widgets/wizard_widgets.dart
LaClimSx, line 1: Not something you changed on this PR, but I notice on line 230 that body can never be null so this check doesn't do anything. Maybe you could remove it ?

LaClimSx (approved):
Thanks for the change!

Thechi2000 (approved):

pmdlt (approved):
lgtm ! Run well on web + android via APK

pmdlt:
@LeanderCretegny seems to be missing code coverage with the recents merges :/

LeanderCretegny:
That is pretty annoying, I checked and the filter is almost completely covered by the tests. It means it is other parts of our code base that is lacking some coverage. I will open an issue that should be merged before this PR.

PR #202, opened 2024-05-31
fix: debounce + state fix for alarm setup view, news settings null fix

### LINKED ISSUE #201: bug: alarm volume slider debouncing
In the alarm view the slider for setting the volume immediately saves the value to the device and Supabase, hence the slow interaction.

For better interaction, the save action must be delayed. Either with a timer or triggered by some unfocus event

loggerz:
Closes #201

The alarm setup view is now simply reactive and glitch-free.

How to test : go to the alarm setup view and fiddle with all the buttons, no incoherent state nor rollback should happen


Pacsort17 (approved):
Still works :) thanks for the fix

LaClimSx (approved):
Tested on emulator, the slider is no longer buggy

PR #203, opened 2024-05-31
170 feat swipe down to regenerate news

LaClimSx:
This PR introduces a new feature : We can now generate a new transcript by swiping down on the news view. This should display a small loading icon, and after some time show the new transcript at the top of the list. Also, when the user doesn't have any news, there isn't a button to generate news anymore, since the user can simply swipe to do it.

You can test this behavior by just launching the app and swiping down to see if it generates a news transcript.


lib/models/offline_recorder.dart
Thechi2000, line 1: You seems to have added carriage returns to the file (because you are using Windows). Please remove them for consistency. You can also change your git settings to automatically correct it to avoid any issue (see [here](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace)).

lib/l10n/app_en.arb
```diff
@@ -47,8 +47,7 @@
   "wizardInterestsTitle": "Select interests",
   "newsLoading": "Please wait while we fetch the news for you",
   "newsEmptyTitle": "You don't have any news yet",
-  "newsEmptyDescription": "The first one will be generated the first time the alarm goes off.",
-  "newsEmptyGenerateButton": "Generate now",
+  "newsEmptyDescription": "The first one will be generated the first time the alarm goes off. You can also generate a transcript by swiping down.",
```
Thechi2000, line 50: I feel like "refreshing the page" would be more intuitive than "swiping down", but I'm not sure about it. What do you think ?

lib/l10n/app_fr.arb
```diff
@@ -35,8 +35,7 @@
     "wizardInterestsTitle": "Sélectionner vos intérêts",
     "newsLoading": "Veuillez patienter pendant que la news est générée",
     "newsEmptyTitle": "Vous n'avez pas encore de news",
-    "newsEmptyDescription": "La première sera générée lorsque votre alarme s'activera",
-    "newsEmptyGenerateButton": "Générer maintenant",
+    "newsEmptyDescription": "La première sera générée lorsque votre alarme s'activera. Vous pouvez également en générer manuellement en défilant vers le bas.",
```
Thechi2000, line 38: Same as above

Thechi2000 (changes requested):

Thechi2000 (approved):
LGTM !

lib/l10n/app_en.arb
loggerz: What I like about the "swiping down" is that it's expressive : you know what to do to refresh. I suggest we keep it that way

Thechi2000 (dismissed):

quartztz (dismissed):
Looks good to me ! Can be merged, nice UX feature

quartztz (approved):
Dismissed because of a merge, but the point still stands ! good feature

Pacsort17 (approved):
I don't know why but my previous review isn't shown in the history, so I'll right it down here.

This is a nice feature ! Tested on Android, it works well and it is userfriendly.
As I told before, a test could have been added for checking that previous news from the same day are still displayed (which is the case currently).

It is also a little bit slow, but it is bottleneck by the LLM so there is not much we can do about it.

PR #205, opened 2024-05-31
Small code ui refactor on share button

Pacsort17:
Put all elements of ui page into a single scaffold
Update share button image

Check : Only one scafold in code, that can be found in masterview
Test : Launch the app and the bottom right button should have a new look


Pacsort17:
I don't know why this PR was created with all ofllineRecorder's commits ^^'

PR #206, opened 2024-06-01
refactor: update "take a picture" button and ui code

Pacsort17:
Put all elements of ui page into a single scaffold
Update "take a photo" button icon

Check : Only one scafold in code, that can be found in masterview
Test : Launch the app and the bottom right button should have a new look (two arrows in a circle) + Take a photo of something that is not a news, and the title should be "An error occured"


lib/views/context_view.dart
```diff
@@ -19,10 +19,12 @@ class ContextView extends StatelessWidget {
...
       return LoadingView(text: loc.ocrLoadingText);
     } else if (nrvm.hasError) {
-      body = ErrorDisplayWidget(description: nrvm.getErrorMessage(loc));
+      body = ErrorDisplayWidget(
+          title: loc.errorTitle, description: nrvm.getErrorMessage(loc));
     } else if (nrvm.result == null) {
       nrvm.takePictureAndProcess();
-      body = ErrorDisplayWidget(description: loc.ocrErrorNoImage);
+      body = ErrorDisplayWidget(
+          title: loc.errorTitle, description: loc.ocrErrorNoImage);
```
Thechi2000, line 27: `loc.errorTitle` is already the default value. Why did you change it to be explicit ?

lib/views/context_view.dart
```diff
@@ -47,17 +49,23 @@ class ContextView extends StatelessWidget {
...
+            body,
+            Align(
+              alignment: Alignment.bottomRight,
+              child: IconButton.filledTonal(
+                iconSize: 40,
+                onPressed: () => nrvm.takePictureAndProcess(),
+                icon: Container(
+                    padding: const EdgeInsets.all(UNIT_PADDING / 4),
+                    child: const Icon(Icons
+                        .sync_outlined)), // Chosen because it represents the act of doing an action again, which here is the case because they have to take a new picture
```
Thechi2000, line 64: Changing the icon is great to avoid confusion with the icon from the navigation bar, however I don't think this one works well. I find that it does not convey the information that the button will allow you to take another picture. `camera` might be more appropriate.

Thechi2000 (changes requested):
The floating button should be handled in the `floatingActionButton` of the scaffold in `MasterView`, as discussed, not in a `Stack` within the `ContextView`. Also, you can apply the same changes to the share button in `NewsView`, since it has the same defect.

lib/views/context_view.dart
Pacsort17: when tested on android, I got a message like "Here is the context :" instead of "An error happened"

lib/views/context_view.dart
Pacsort17: I thought also of camera, but it is still used ? I haven't seen this icon since a while, and I don't think users while guess that it is a camera

Thechi2000: It's true that I haven't seen the `camera` icon in a while, but it is still more clear than the `sync_outlined`. For me, the latter really is a "reload" kind of icon, which would refresh the output of the scan.

pmdlt: I prefer camera too

LeanderCretegny: Personally I think the icon that we should change is the one in the bottom bar and not the one inside the context view. Unfortunately I (for now) don't know which icon we should use

lib/views/context_view.dart
Pacsort17: Tested after running flutter clean, it didn't happen again

Thechi2000 (dismissed):
Nice, thanks for the changes !

lib/views/master_view.dart
```diff
@@ -45,15 +52,58 @@ class _MasterView extends State<MasterView> {
...
+        floatingButton = IconButton.filledTonal(
+          iconSize: 40,
+          onPressed: () =>
+              Provider.of<NewsRecognitionViewModel>(context, listen: false)
+                  .takePictureAndProcess(),
+          icon: Container(
+              padding: const EdgeInsets.all(UNIT_PADDING / 4),
+              child: const Icon(Icons.camera_alt)),
+          color: THEME_BUTTON,
+        );
```
LaClimSx, line 106: I think it would look better if this wasn't a filledTonal button, since it always shows the overlay.

LaClimSx (changes requested):
I left a comment about the UI of the floating button

lib/views/master_view.dart
loggerz: The bare icon looks strange, it doesn't look like a button in that case. Maybe another design ?

LaClimSx: Maybe you could use a button that looks like the one we have on the news view (the share button)

LaClimSx: Yes, maybe we could use a design like the one on the news view (the share button)

loggerz:
Just waiting for the button change, the rest is good !

lib/views/master_view.dart
Pacsort17: :'(

loggerz (approved):
Nice, and with a better animation than the first view !

LaClimSx (approved):
Thanks for the change, the code works well otherwise

PR #209, opened 2024-06-01
feat: Context history

LaClimSx:
This PR changes the way the context view works, so that every time you take a picture of some text, it stores it in a file, and every time you load this view, it displays all the previous contexts you have asked for.

I couldn't test this on my emualtor so I'm curious to see the behavior on an actual phone : What happens if you completely close the app and open it again ?


lib/viewmodels/news_recognition.dart
```diff
@@ -36,6 +43,7 @@ class NewsRecognitionViewModel extends ChangeNotifier {
   }
 
   @protected
+  // ignore: library_private_types_in_public_api
```
Thechi2000, line 46: This does not seem to be needed anymore.

lib/viewmodels/news_recognition.dart
```diff
@@ -118,10 +127,46 @@ class NewsRecognitionViewModel extends ChangeNotifier {
     markProcessingAs(true);
     final text = await recognizeText(picture.path);
     await invokeProcessImage(text);
+    // Write the context to a file
+    await writeContext(result);
```
Thechi2000, line 131: This could be ran as a background task, if it is not to complex to implement, thus the user could have the result immediately.

Thechi2000 (commented):
Good work !

Just a small potential issue: when restarting the app and going to the context view, we get the "no image taken" error, and we cannot see the previous contexts without taking a new picture. Shouldn't we get rid of that error and simply display the previous contexts ?

lib/views/context_view.dart
```diff
@@ -24,6 +26,41 @@ class ContextView extends StatelessWidget {
...
+                      indent: UNIT_PADDING * 3,
+                      endIndent: UNIT_PADDING * 3,
+                      color: THEME_GREY,
+                    )
+                  // We do +1 because we don't want to display the last context, since it is already in the result
+                  : Text(nrvm.contexts[(index / 2).floor() + 1],
+                      style: Theme.of(context).textTheme.displaySmall,
+                      textAlign: TextAlign.center);
+            },
+          ));
```
LeanderCretegny, line 63: Please use a `Column()` instead of `ListView()` with `NeverScrollableScrollPhysics()`, it will make the code way simpler and more readable.

LeanderCretegny (changes requested):
Good code overall, my only remarks is about a miss used `ListView()`. The refactor on floating buttons color is a good idea!

lib/viewmodels/news_recognition.dart
LaClimSx: Removed it, thanks

lib/viewmodels/news_recognition.dart
LaClimSx: Good idea, I changed it

lib/views/context_view.dart
LaClimSx: I tried using a column but since we are using tricky arithmetic to know which widgets to add to the listView, I feel that using a column would be more complicated, especially since we would need to have nested columns

quartztz (approved):
Could use a little more documentation, but it looks good as is ! 
Thank you for the much needed feature, let's merge this :D

LeanderCretegny (approved):

PR #211, opened 2024-06-02
feat: implement new alarm view 

quartztz:
Implements a new view following more "usual" standards. 

Closes an issue !


lib/views/news_alert_setup_view.dart
```diff
@@ -86,141 +87,132 @@ class _NewsAlertSetupViewState extends State<NewsAlertSetupView> {
...
-                        ),
-                        Switch(
+    Widget body = Container(
+        padding: const EdgeInsets.symmetric(horizontal: UNIT_PADDING * 3),
+        child: ListView(
+          shrinkWrap: true,
+          children: <Widget>[
+            Container(
+                padding: const EdgeInsets.only(
+                    top: UNIT_PADDING * 8, bottom: UNIT_PADDING * 2),
```
LaClimSx, line 97: I think it is maybe too much padding, it looks like this on my small emulator, so there is a big white space and I need to swipe to see the whole view.
Capture d'écran 2024-06-02 170840

Maybe you could remove some padding so it looks like this ?
Capture d'écran 2024-06-02 171042

LaClimSx (changes requested):
Nice UI refactor, I've added a comment about padding at the top of the view

LaClimSx (approved):
Thanks for the changes, the rest looks good!

Pacsort17 (approved):
Tested on Android, it works weell, thansk for the work :)

PR #212, opened 2024-06-02
Revert "feat: replace guest login with anonymous sign in option"

pmdlt:
Reverts ActualIA/app#189

PR #213, opened 2024-06-02
Replace guest login with anonymous sign in option

pmdlt:
This PR replaces #189 that was initially merged and reverted (#212) (following a pending commit).
All the work goes to @quartztz

Original text:

This pull request will modify the codebase's approach to anonymous/temporary users by using the existing [Anonymous Sign-In API](https://supabase.com/blog/anonymous-sign-ins) that's made available by Supabase Auth. This is a cleaner solution, as it allows users to use the app as intended without a Google Sign-In, with their own interests and so on, without impacting others in the same situation. 

Still in drafts !

A few things must be noted. 
> ℹ️  I had to modify the settings in the Supabase dashboard to explicitly allow Anonymous Sign-Ins. Be mindful of that !

> ⚠️ This inevitably opens us to security concerns that we didn't have earlier, such as potential botting practices or other exotic attacks through anonymous users. We might want to see how this interacts with our RLS, or if we want to add some sort of Captcha anti-bot behavior.


LaClimSx (approved):

quartztz (approved):
LGTM ! go go go go go !

